PR #5, opened 2022-11-02
add mock category relation

MagicTearsAsunder:


annotation/app/categories/resources.py
```diff
@@ -104,7 +106,7 @@ def get_child_categories(
 @router.post(
     "/search",
     status_code=status.HTTP_200_OK,
-    response_model=Page[Any],  # type: ignore
+    response_model=Page[Any],
```
khyurri, line 109: `Any` might be changed to pydantic model, which will be describe response schema. Right now it's not clear, what will be returned

khyurri (commented):
Thank you for PR
Please, take a look into comments

khyurri:
@MagicTearsAsunder 
May I ask you to PR in different branch, because this PR is first step for hierarchical labels

khyurri:
@MagicTearsAsunder 
As we discussed, this PR also should add hierarchical response into `/jobs/{job_id}/categories/`
May I ask you to check other GET responses, maybe we forgot some places?

PR #6, opened 2022-11-04
ci/cd: fix env vars, add parameters to charts and docker files

theoriginmm:
Manual Deployment. Will be replaced soon.


khyurri (commented):
@theoriginmm 

Thank you for PR!

Don't you want to move registry host also to variables? 

Also, I can see dependencies from internal EPAM artifactory, should we move it into ECR?

theoriginmm:
> @theoriginmm
> 
> Thank you for PR!
> 
> Don't you want to move registry host also to variables?
> 
> Also, I can see dependencies from internal EPAM artifactory, should we move it into ECR?

@khyurri  Thank you for the comment.
This  PR simplifies manual deployment, eliminates artifactory url from default variables.

We need a default value  as image name for makefiles.

Yes,  it's possible. Moved to ECR.

khyurri (approved):

PR #19, opened 2022-11-14
Feature: LTREE supported operations#2

### LINKED ISSUE #2: Annotation service: add hierarchy support to categories

minev-dev:
* Add support of LTREE (Postgre-only) operations:
  * "parent" - return parent of record with provided id
  * "parents_recursive" - return all ancestors of record with provided id
  * "children" - get first-level children for record with provided id
  * "children_recursive" - get all descendants for record with provided id
* Add tests to cover this operations. Tests are checking sql statements for now because SQLite doesn't support LTREE data type
* Minor style fixes

### DIFF
--- a/lib/filter_lib/README.md
+++ b/lib/filter_lib/README.md
@@ -242,6 +242,10 @@ Available filter operations:
 - ``any``
 - ``not_any``
 - ``distinct``
+- ``parent``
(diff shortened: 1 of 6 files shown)


khyurri (approved):
LGFM

PR #23, opened 2022-11-15
refactor: add compatibility with filter_lib

MagicTearsAsunder:
#2


annotation/app/categories/resources.py
```diff
@@ -107,14 +104,14 @@ def get_child_categories(
 @router.post(
     "/search",
     status_code=status.HTTP_200_OK,
-    response_model=Page[CategoryResponseSchema],
+    response_model=Page[Union[CategoryResponseSchema, str, dict]],
```
khyurri, line 107: In this level better to cast str and dict to `CategoryResponseSchema`

You can create additional casting function in exceptional cases

khyurri: We discussed, that this behaviour required for front-end

Issue #13, opened 2022-11-09
Document labelling: support of multi-label tagging on document level (e.g. for document classification purposes).


khyurri:
Back-end:
- [x] Add labels for each revision, not in manifest level

Design:
- [x] Design how to set tag per document

Front-end:
- [x] Implement

Kitonick79:
This task is about classifying document with multiple labels. We need to get model output, take labels for a document as a whole and display it on UI. We would need some UX support here.

MikhailFokanov:
As we discussed today to save the time and effort, the acceptable and solid workaround is to have a simple category named "bucket". The annotator will draw a rectangle in the free sapce in the top right corner with "free-box" tool and then select the right entity from taxonomy in the data tab (as usual). No changes should be done on UI in such case. On the export process converter can just find the box with bucket category and set it to the document.
Multiple labels are also available out-of-the-box with such approach.

Kitonick79:
> 
But how should it be imported from model's output?

MikhailFokanov:
Nothing should be done on the backend/frontend side, if the proposed solution with box is accepted by customer. Only during the export process converter can just find the box with bucket category and set it to the document.

khyurri:
@MikhailFokanov 
Could you please explain what is box and what is export process? 

This task is about to add users functionality to assign category to whole document. This is exactly same, as assign category to some bbox, but to whole document. 

Please, check PR description, @cakeinsauce  already added API and we wanted to finish from front-end task

khyurri:
FE is required

Iogsotot:
Need to fix BE https://github.com/epam/badgerdoc/issues/128

Iogsotot:
https://github.com/epam/badgerdoc/pull/157 done

PR #31, opened 2022-11-17
feat: add labeling for documents

cakeinsauce:
Issue: #13 

I have adjusted the document schema in DB by adding a new column `category_id`  (`annotated_docs`  table) which represents corresponding category's pk in the `categories` table. Categories have recursive structure (`LTree`) so we can use them for multi-labeling of documents as well.

- [x] Documents support labeling
- [ ] Covered by tests


annotation/app/models.py
```diff
@@ -91,6 +91,11 @@ class AnnotatedDoc(Base):
     validated = Column(ARRAY(INTEGER), nullable=False, server_default="{}")
     tenant = Column(VARCHAR, nullable=False)
     task_id = Column(INTEGER, ForeignKey("tasks.id", ondelete="SET NULL"))
+    category_id = Column(
```
khyurri, line 94: I think better to add many to many support and allow users choose many labels per document

cakeinsauce: But don't we have a tree-like structure of categories? So we could add labels like this? Or I miss something?

khyurri: I mean, user may choose categories from different subtrees and we might support this functionality

khyurri (approved):
LGFM

PR #32, opened 2022-11-17
fix: special symbols in category id

MagicTearsAsunder:
Issue: #2

Fixed 500 Internal server error on creating `Category` with _id_ containing special chars such as "-", "!", "." etc.
Added validator, now only alphanumeric are allowed.
Added tests for such cases and also fixed existed ones.

Edit:
Allowed underscore "_".
Replaced 'parents' list attribute with the 'has_children' bool value in Category response.
Updated openapi.yaml

### DIFF
--- a/annotation/app/categories/services.py
+++ b/annotation/app/categories/services.py
@@ -49,7 +49,7 @@ def add_category_db(
         tree = Ltree(f'{category_input.id}')
 
     category = Category(
-        id=(id_ or str(uuid.uuid4())),
+        id=(id_ or str(uuid.uuid4()).replace('-', '')),
         name=name,
         tenant=tenant,
         parent=parent if parent != "null" else None,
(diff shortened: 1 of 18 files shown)


annotation/app/categories/services.py
```diff
@@ -49,7 +49,7 @@ def add_category_db(
         tree = Ltree(f'{category_input.id}')
 
     category = Category(
-        id=(id_ or str(uuid.uuid4())),
+        id=(id_ or str(uuid.uuid4()).replace('-', '')),
```
khyurri, line 52: just use `uuid.uuid4().hex`

annotation/app/schemas/categories.py
```diff
@@ -28,6 +29,12 @@ class CategoryInputSchema(CategoryBaseSchema):
         description="If id is not provided, generates it as a UUID.",
     )
 
+    @validator('id')
+    def alphanumeric_validator(cls, value):
```
khyurri, line 33: Add support `_`

Issue #57, opened 2022-12-01
Change Category -> Taxonomy relation from OneToOne to ManyToMany


MagicTearsAsunder:
There is existing requests link between microservices _annotation_ (Category model) and _taxonomy_ (Taxonomy model). Change its relation from OneToOne to ManyToMany

khyurri:
Released in 1.1.0

PR #61, opened 2022-12-02
feat: change taxonomy category relation into many to many

MagicTearsAsunder:
Issue #57


khyurri (approved):
Approved, because blocker. Didn't validate code

PR #71, opened 2022-12-07
feat: documents pair classification

### LINKED ISSUE #14: Document labelling: document pair classification (for example, determining if the texts are similar or not)
## Front-end 

- [x] Design is required

## Front-end 

- [x] Add view to switch between document labels and document links
- [x] Add component which shows document links(categories) in accordion
- [x] Linked documents should be shown under selected category
- [x] Show documents side-by-side when selecting linked document
- [x] Add an ability to select relation(category) between shown documents
- [x] Add api calls to get and send selected links

## Back-end

- [x] Have to support relations between documents: similar, identical, etc. One document may have multiple relations. Back-end must support different types of relationships
- [x] Add linking between documents by link type. Link and documents must be in revision
- [x] Add support (or check support) of initial load, when we upload revision for all annotators (check with Mikhail)

## General algorithm
Back-end must support tasks to check pairs.

cakeinsauce:
Closes #14 
- [x] Documents support pair classification for:
  - [x] Saving
  - [x] Retrieving
- [x] Covered by tests
- [x] Alembic migrations

### DIFF
--- a/annotation/app/annotations/main.py
+++ b/annotation/app/annotations/main.py
@@ -2,7 +2,7 @@ import json
 import os
 from datetime import datetime
 from hashlib import sha1
-from typing import Any, Dict, List, Optional, Set, Tuple, Union
+from typing import Dict, List, Optional, Set, Tuple, Union
 from uuid import UUID
 
 import boto3
@@ -213,7 +213,6 @@ def create_manifest_json(
     bucket_name: str,
     job_id: int,
     file_id: int,
-    doc_categories: Optional[List[str]],
     db: Session,
     s3_resource: boto3.resource,
 ) -> None:
@@ -247,7 +246,6 @@ def create_manifest_json(
     manifest["failed_validation_pages"] = list(failed_validation)
     manifest["file"] = s3_file_path
     manifest["bucket"] = s3_file_bucket
-    manifest["categories"] = doc_categories
 
     manifest_json = json.dumps(manifest)
     upload_json_to_minio(
@@ -314,6 +312,14 @@ def construct_annotated_doc(
         or not. If flag is False, merge conflict will occur
     :return: newly created revision or latest revision
     """
+    from app.categories import fetch_bunch_categories_db
+
+    doc_categories = fetch_bunch_categories_db(
+        db=db,
+        category_ids=set(doc.categories or []),
+        tenant=tenant,
+        root_parents=True,
+    )
     annotated_doc = AnnotatedDoc(
         user=user_id,
         pipeline=pipeline_id,
@@ -323,6 +329,7 @@ def construct_annotated_doc(
         failed_validation_pages=doc.failed_validation_pages,
         tenant=tenant,
         task_id=task_id,
+        categories=doc_categories,
     )
     s3_path = f"{S3_START_PATH}/{str(job_id)}/{str(file_id)}"
 
@@ -391,7 +398,6 @@ def construct_annotated_doc(
         bucket_name,
         job_id,
         file_id,
-        doc.categories,
         db,
         s3_resource,
     )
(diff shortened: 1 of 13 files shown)


annotation/app/annotations/main.py
```diff
@@ -358,8 +360,18 @@ def construct_annotated_doc(
...
+    document_links = construct_document_links(
+        annotated_doc, doc.similar_revisions or []
+    )
     db.add(annotated_doc)
-    db.flush()
+    db.add_all(document_links)
+    try:
+        db.flush()  # TODO: shouldn't we commit here?
+    except IntegrityError as err:
+        logger_.exception(
```
andrei-shulaev, line 371: rollback changes in case of error.

annotation/app/annotations/main.py
```diff
@@ -358,8 +360,18 @@ def construct_annotated_doc(
...
         return latest_doc
 
+    document_links = construct_document_links(
+        annotated_doc, doc.similar_revisions or []
+    )
     db.add(annotated_doc)
-    db.flush()
+    db.add_all(document_links)
+    try:
+        db.flush()  # TODO: shouldn't we commit here?
```
andrei-shulaev, line 369: if we do db.commit() later, no need to flush here. flush just moves data from python transaction buffer to DB transaction buffer

annotation/app/models.py
```diff
@@ -332,3 +359,51 @@ def get_stats(self):
...
+class DocumentLinks(Base):
+    __tablename__ = "document_links"
+
+    original_revision = Column(VARCHAR)
+    original_file_id = Column(INTEGER)
+    original_job_id = Column(INTEGER)
+
+    similar_revision = Column(VARCHAR)
+    similar_file_id = Column(INTEGER)
+    similar_job_id = Column(INTEGER)
```
andrei-shulaev, line 373: Can we make this as composite key?

annotation/app/models.py
```diff
@@ -332,3 +359,51 @@ def get_stats(self):
...
+
+    label = Column(VARCHAR, nullable=False)
+
+    __table_args__ = (
+        ForeignKeyConstraint(
+            (original_revision, original_file_id, original_job_id),
+            (AnnotatedDoc.revision, AnnotatedDoc.file_id, AnnotatedDoc.job_id),
+        ),
+        PrimaryKeyConstraint(
```
andrei-shulaev, line 383: We need to add similar doc as primary key as well. Since it is possible to have one to many link here.

andrei-shulaev (changes requested):

annotation/app/annotations/main.py
cakeinsauce: Good catch, added.

annotation/app/annotations/main.py
cakeinsauce: Brief check didn't give the answer where do we commit this transaction =)
Changed to commit

annotation/app/models.py
cakeinsauce: You are right, missed this. Added similar doc as primary key+

annotation/app/models.py
cakeinsauce: I actually do [here](https://github.com/epam/badgerdoc/pull/71/files/ee263455b1a742f1789eaaf23364ffa7faa1fe39#diff-3ac0632a865326abdf7db8425d85b915e0ce5fcff9516381b0cfc6e7a276a036R382-R384). SQLAlchemy magic =) 
Because these columns are FKs to a composite key as well we have to mark them like I did, otherwise SQLAlchemy assumes we use only one column referring to a composite key, which leads us to complains about PK constraint in `annotated_doc` by ORM

andrei-shulaev (approved):

Issue #64, opened 2022-12-06
Progress report per annotator (doc name, annotator id, date of finishing)


khyurri:
## General algorithm

User will be able to get report by period. Report must contain:

By annotator:
- Finished tasks + AS + (time_start, time_finish to calculate how much time annotator worked in single document)
- Tasks in progress + job start time

Example: 

`annotator_id, task_id, task_status, time_start (earliest event), time_finish (draft or task_finish), agreement score (if task finished and job with extensive coverage and AGREEMENT_SCORE_SERVICE_URL was set)` 

Front-end:
- [x] send event (post) when document is open by annotator or validator during task
- [x] add page to download result by user and date_from, date_to

Back-end:
- [x] add endpoint to save document open timestamp, we need to create new table: `annotation_statistics` with fields: `job_id, task_id, event_type, event_date (current timestamp), additional_data (json)`. event_type must be: `document_opened`
- [x] add new config value (annotation service): `AGREEMENT_SCORE_SERVICE_URL`. In case of extensive coverage task, when validator finishes validation task, create additional POST request to `AGREEMENT_SCORE_SERVICE_URL` (if set) to get agreement score. BadgerDoc must send: annotators_id, job_id, task_id, document url (S3 resource), manifests urls (S3 resource) for each annotator
- [x] Store all metrics between annotator and validator. Each annotator has own AS for each task over validator.

MagicTearsAsunder:
`AGREEMENT_SCORE_SERVICE_URL`:
POST REQUEST
Body
```
[
    {
        "annotator_id": "f0474853-f733-41c0-b897-90b788b822e3",
        "job_id": 1,
        "task_id": 1,
        "s3_file_path": "file1.pdf",
        "s3_file_bucket": "s3://bucket",
        "manifest_url": "url1",
    },
    {
        "annotator_id": "4e9c5839-f63b-49c8-b918-614b87813e53",
        "job_id": 1,
        "task_id": 1,
        "s3_file_path": "file2.pdf",
        "s3_file_bucket": "s3://bucket",
        "manifest_url": "url2",
    }
]
```

EXPECTED RESPONSE:
Body
```
[
    {
        "annotator_id": "f0474853-f733-41c0-b897-90b788b822e3",
        "job_id": 1,
        "task_id": 1,
        "agreement_score": {"attr1": "value1"}
    },
    {
        "annotator_id": "4e9c5839-f63b-49c8-b918-614b87813e53",
        "job_id": 1,
[...]

PR #81, opened 2022-12-12
feat: add progress report per annotator

MagicTearsAsunder:
Issue: #64 

- Added endpoint to save document stats open/close.
- Add request to the external service to get Agreement Score.
- Add table to save all metrics between annotator and validator.
- Add Agreement Score stats export feature.

- Fixed bug category search/category delete on Ltree is null.
- Fixed Category id special chars and allow only alphanumeric and underscore.

TODO: add tests related to the Agreement Score


web/src/pages/reports/reports-page.tsx
```diff
@@ -0,0 +1,12 @@
+import { ReportsConnector } from 'connectors/reports-connector';
+import React from 'react';
+
+const ReportsPage = () => {
+    return (
+        <>
```
Iogsotot, line 6: It's not necessary to use fragment here

Iogsotot (approved):

iurii-topychkanov (approved):

annotation/app/errors.py
```diff
@@ -51,6 +51,11 @@ def __init__(self, exc_info: Union[str, RequestException]):
         self.exc_info = exc_info
 
 
+class AgreementScoreServiceException(Exception):
+    def __init__(self, exc: Union[str, RequestException]):
```
cakeinsauce, line 55: Is it okay behavior? I believe you should receive message only (`str`). Or if you want to save error type as well, you may want do this
```python
def __init__(self, message, errors):            
        # Call the base class constructor with the parameters it needs
        super().__init__(message)
            
        # Now for your custom code...
        self.errors = errors
```

annotation/app/microservice_communication/task.py
```diff
@@ -0,0 +1,46 @@
...
+                HEADER_TENANT: tenant,
+                AUTHORIZATION: f"{BEARER} {token}",
+            },
+            json=body,
+            timeout=5,
+        )
+        if response.status_code != 200:
+            raise AgreementScoreServiceException(response.text)
+    except RequestException as exc:
+        raise AgreementScoreServiceException(exc)
```
cakeinsauce, line 42: You should use your custom message or get it from the `exc`. And to save traceback use `raise from` syntax
```suggestion
        raise AgreementScoreServiceException(str(exc)) from exc
```

annotation/app/models.py
```diff
@@ -97,6 +102,13 @@ class AnnotatedDoc(Base):
     tasks = relationship("ManualAnnotationTask", back_populates="docs")
 
 
+def default_tree(column_name: str) -> Callable:
```
cakeinsauce, line 105: May you specify callable args and return type? Strange that mypy doesn't complain about it

annotation/app/models.py
```diff
@@ -97,6 +102,13 @@ class AnnotatedDoc(Base):
...
+    def default_function(context) -> str:
+        return Ltree(f"{context.current_parameters.get(column_name)}")
```
cakeinsauce, line 107: If we get None, should it be "None"(str)?

annotation/app/models.py
```diff
@@ -244,3 +264,69 @@ class ManualAnnotationTask(Base):
...
+        }
+
+
+class AgreementScore(Base):
+    __tablename__ = "agreement_score"
+    annotator_id = Column(
+        UUID(as_uuid=True),
+        ForeignKey("users.user_id"),
+        nullable=False,
+        primary_key=True,
```
cakeinsauce, line 307: So agreement score is only one per annorator?
And I believe you may omit `nullable=False` since primary key is treated by sqlalchemy as so

annotation/app/tasks/services.py
```diff
@@ -381,3 +389,109 @@ def unblock_validation_tasks(db: Session, task: ManualAnnotationTask) -> None:
...
+    db: Session,
+    task_id: int,
+    stats: AnnotationStatisticsInputSchema,
+) -> AnnotationStatistics:
+    stats_db = get_task_stats_by_id(db, task_id)
+
+    if stats_db:
+        for name, value in stats.dict().items():
+            setattr(stats_db, name, value)
+        stats_db.updated = datetime.now()
```
cakeinsauce, line 415:
```suggestion
        stats_db.updated = datetime.utcnow()
```

cakeinsauce (changes requested):

annotation/app/models.py
MagicTearsAsunder: default tree depends on default id which can not be not. But yes, this is an implicit behaviour

annotation/app/models.py
MagicTearsAsunder: 1st arg can not be specified because it can not be imported

annotation/app/models.py
MagicTearsAsunder: Yes, thats wrong. One score per task, of course.

annotation/app/tasks/services.py
MagicTearsAsunder: done as suggested, also required to change model defaults

cakeinsauce (approved):
LGTM! May you squash commits before merging?

PR #82, opened 2022-12-12
CI fix postgres and enable all tests

thinklab:


andrejshulaev:
Let's add something like this https://github.com/marketplace/actions/pytest-coverage-comment ? may be not in current PR, but would be nice to have it.

PR #86, opened 2022-12-13
Feature/documents filtering by dataset#59

Iogsotot:
Fixed bug with getting datasets when filtering
image

### DIFF
--- a/annotation/app/categories/resources.py
+++ b/annotation/app/categories/resources.py
@@ -82,7 +82,9 @@ def fetch_category(
     x_current_tenant: str = X_CURRENT_TENANT_HEADER,
 ) -> CategoryResponseSchema:
     category_db = fetch_category_db(db, category_id, x_current_tenant)
-    category_response = insert_category_tree(db, category_db)
+    category_response = insert_category_tree(
+        db, category_db, tenant=x_current_tenant
+    )
     return category_response
 
 
--- a/annotation/app/categories/services.py
+++ b/annotation/app/categories/services.py
@@ -33,19 +33,43 @@ cache = TTLCache(maxsize=128, ttl=300)
 logger = app_logger.Logger
 
 
+def is_category_leaf(db: Session, category: Category, tenant: str) -> bool:
+    return not (
+        db.query(Category.id)
+        .filter(
+            and_(
+                Category.parent == category.id,
+                or_(Category.tenant == tenant, Category.tenant == null()),
+            )
+        )
+        .first()
+    )
+
+
+def set_parents_is_leaf(
+    category_db: Category,
+    parents: Optional[List[CategoryResponseSchema]] = None,
+    is_leaf: bool = False,
+) -> CategoryResponseSchema:
+    if parents is None:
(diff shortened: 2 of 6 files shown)


web/src/api/hooks/assets.ts
```diff
@@ -104,6 +104,18 @@ export const fetchThumbnailPiece = async (
     }
 };
 
+// TODO: delete if search api gets fix (dataset.id filters)
```
iurii-topychkanov, line 107: Can we remove it now?

web/src/connectors/documents-sidebar-connector/documents-sidebar-connector.tsx
```diff
@@ -6,7 +6,7 @@ import styles from './documents-sidebar-connector.module.scss';
 import {
     Filter,
     PagedResponse,
-    Pipeline,
+    // Pipeline,
```
iurii-topychkanov, line 9: Could be removed

web/src/connectors/documents-sidebar-connector/documents-sidebar-connector.tsx
```diff
@@ -41,7 +41,7 @@ export const DocumentsSidebarConnector = <T extends object>({
     filters
 }: DocumentsSidebarConnectorProps<T>) => {
     const [searchText, setSearchText] = useState('');
-    const [sortConfig, setSortConfig] = useState({
+    const [sortConfig] = useState({
```
iurii-topychkanov, line 44: If we are going to use it this way, you'll actually create immutable `sortConfig`, since there will be no way to update it. 
Since `sortConfig` relies on `sortField` probably it's better to store it in constant e.g `const sortConfig = {...}` in that case if component will be re-rendered and `sortField` will be updated, the sortConfig will be updated in accordance with the changes

web/src/connectors/documents-table-connector/documents-table-connector.tsx
```diff
@@ -105,14 +106,19 @@ export const DocumentsTableConnector: React.FC<DocumentsTableConnectorProps> = (
             onFilesSelect(selectedFiles as number[]);
         }
     }, [selectedFiles]);
+    // TODO: check if this can be removed
```
iurii-topychkanov, line 109: can be removed

web/src/connectors/documents-table-connector/documents-table-connector.tsx
```diff
@@ -186,11 +205,11 @@ export const DocumentsTableConnector: React.FC<DocumentsTableConnectorProps> = (
         sortConfig,
         filters
     );
-
-    const getCardSort = () => ({
-        direction: 'desc',
-        field: documentsSort
-    });
+    // TODO: check if this can be removed
+    // const getCardSort = () => ({
```
iurii-topychkanov, line 209: can be removed?

Iogsotot: @sbilevich can we?

PR #99, opened 2022-12-19
Issue 91 - fix download thumbnail bug

spiridonovfed:


spiridonovfed:
- [ ] tested on dev env

PR #106, opened 2022-12-19
test: fix tests related to the extensive coverage

MagicTearsAsunder:


annotation/tests/test_post_job.py
```diff
@@ -763,7 +757,6 @@ def test_post_extraction_job_saved(
     assert saved_job.get("job_type") == JobTypeEnumSchema.ExtractionJob
 
 
-@pytest.mark.integration
```
andrejshulaev, line 766: Leave it as is, since we still use DB in this test.

andrejshulaev (changes requested):

andrejshulaev (approved):
LGTM

Issue #62, opened 2022-12-06
Integration of custom agreement metrics on the batch level (to assess the overall level of agreement between annotators) 


khyurri:
## General algorithm

Example: 2 annotators annotates 1 document, if they annotation is not match, we need to have algorithm to compare annotations if they not match. Agreement metrics helps to check if annotations are matched. 

- Add config value `AGREEMENT_SCORE_MIN_MATCH` `float` 
- When job task transitions to validation state, BD makes request to external service for getting is validation required
- Metrics must be stored in task level into database

This task is very close to: https://github.com/epam/badgerdoc/issues/64

Table agreement_scores:

Tasks: 1, 2, 3 <-- one part of document 
task_from, task_to, agreement_score
1, 2, 0.95
2, 1, 0.95 <--- doesn't make sense 
2, 3, 0.97
1, 3, 0.98

PR #138, opened 2022-12-28
feat: agreement score for multiple annotators

Nathicanaa:
PR for issue #62 

- Added config value `AGREEMENT_SCORE_MIN_MATCH`, type FLOAT
- Added call to `agreement score external service` if all annotation tasks were finished. Note: a code snippet that calls `agreement score external service` is copy-pasted from #64 
- Implemented function for creating metrics from `agreement score external service` response
- Created new table `agreement_metrics` for storing metrics
- Implemented logic for extension coverage validation type: if all annotation tasks reached the `AGREEMENT_SCORE_MIN_MATCH` then auto update validation task with status `finished` and update job status respectively.
- New features are covered with tests.


annotation/app/models.py
```diff
@@ -419,3 +420,21 @@ def __repr__(self) -> str:
...
+        INTEGER,
+        ForeignKey("tasks.id", ondelete="cascade"),
+        primary_key=True,
+        unique=False,
+    )
+    task_to = Column(
+        INTEGER,
+        ForeignKey("tasks.id", ondelete="cascade"),
+        primary_key=True,
+        unique=False,
```
cakeinsauce, line 438: May you clarify why we mark these as `primary_key=True` and `unique=False` for both columns? I am a bit unfamiliar with this: what's the purpose of this? What if we get agreement score for same tasks multiple times? It will erase the old one? So we only have only one agreement score at the same time?

annotation/alembic/versions/7cc1ed83c309_compare_agreement_scores.py
```diff
@@ -0,0 +1,38 @@
...
+    op.create_table(
+        "agreement_metrics",
+        sa.Column("task_from", sa.INTEGER(), nullable=False),
+        sa.Column("task_to", sa.INTEGER(), nullable=False),
+        sa.Column("agreement_metric", sa.FLOAT(), nullable=True),
+        sa.ForeignKeyConstraint(
+            ["task_from"], ["tasks.id"], ondelete="cascade"
+        ),
+        sa.ForeignKeyConstraint(["task_to"], ["tasks.id"], ondelete="cascade"),
+        sa.PrimaryKeyConstraint("task_from", "task_to"),
```
cakeinsauce, line 30: May you adjust migration to the corresponding ORM model in `models.py`? More specifically `agreement_metric`is nullable here, but not in the model. I believe we don't need a record with null metric.
And may you specify `unique=False` here as well?

annotation/app/tasks/resources.py
```diff
@@ -949,6 +957,91 @@ def finish_task(
                 detail=f"Error: connection error ({exc.exc_info})",
             )
         update_inner_job_status(db, task.job_id, JobStatusEnumSchema.finished)
+
+    elif (
+        job.validation_type == ValidationSchema.extensive_coverage
+        and os.getenv("AGREEMENT_SCORE_ENABLED", "false") == "true"
```
cakeinsauce, line 963: May we move it to constant var as you did on line 99 for agreement score?

annotation/app/tasks/resources.py
```diff
@@ -949,6 +957,91 @@ def finish_task(
...
+    ):
+        annotation_tasks: List[ManualAnnotationTask] = (
+            db.query(ManualAnnotationTask)
+            .filter(
+                ManualAnnotationTask.job_id == task.job_id,
+                ManualAnnotationTask.file_id == task.file_id,
+                ManualAnnotationTask.is_validation.is_(False),
+            )
+            .count()
+        )
```
cakeinsauce, line 973: May you move this DB interaction to `services.py` to separate this logic from controller?

annotation/app/tasks/resources.py
```diff
@@ -949,6 +957,91 @@ def finish_task(
...
+                db.query(ManualAnnotationTask).filter(
+                    ManualAnnotationTask.job_id == task.job_id,
+                    ManualAnnotationTask.file_id == task.file_id,
+                    ManualAnnotationTask.is_validation.is_(True),
+                ).with_for_update().update(
+                    {
+                        ManualAnnotationTask.status: TaskStatusEnumSchema.finished  # noqa: E501
+                    },
+                    synchronize_session="fetch",
+                )
```
cakeinsauce, line 1018: May you move this to `services.py` as well?

annotation/app/tasks/resources.py
```diff
@@ -949,6 +957,91 @@ def finish_task(
...
+                    task_from=el.task_from_id,
+                    task_to=el.task_to_id,
+                    agreement_metric=el.metric_score,
+                )
+                for el in compared_score.task_metrics
+            ]
+            db.bulk_save_objects(metrics)
     db.flush()
     update_user_overall_load(db, task.user_id)
     db.commit()
```
cakeinsauce, line 1047: May you move this to `services.py` as well?

annotation/app/tasks/services.py
```diff
@@ -487,3 +491,59 @@ def create_export_csv(
...
+            el = _MetricScoreTuple(
+                task_from_id, score.task_id, score.agreement_score
+            )
+            unique_scores.add(
+                _MetricScoreTuple(
+                    min(el.task_from, el.task_to),
+                    max(el.task_from, el.task_to),
+                    el.score,
+                )
+            )
```
cakeinsauce, line 529: May you move it to another function, like `get_unique_scores(scores: List[ResponseScore], task_id: int) -> Set[_MetricScoreTuple]` to make this function a bit less "god-like" and increase readability. Optional

annotation/app/models.py
Nathicanaa: here is an example of agreement metric for 3 annotators, i believe it expands why do we need `unique=False` for those columns:

task_from, task_to, agreement_score
1, 2, 0.95
1, 3, 0.98
2, 3, 0.97

If i understand business requirements correctly there is no case for getting agreement score for same tasks multiple times - agreement score external service must be called exact once for task group. I believe in potential case of `re-score` metrics we should implement API with UPDATE functionality.

To be honest  i have no clue should we avoid `primary_key=True` for those columns - we need a primary key anyways so we can add column like `id` but i think it will be useless. I guess it's an open question.

annotation/alembic/versions/7cc1ed83c309_compare_agreement_scores.py
Nathicanaa: Actually `agreement_metric` was nullable in models and also was nullable here.  But i agree we don't need this column nullable so it's fixed.
`unique=False` it is not an alembic syntax. if you want to specify unique column into alembic migration you should use `sqlalchemy.UniqueConstraint`, since we don't use `UniqueConstraint` here those columns aren't unique.

annotation/app/tasks/resources.py
Nathicanaa: it is a dummy temporary solution to make this block of code unreachable until external `agreement score service` implemented. Once it's done we should just remove second condition `and os.getenv("AGREEMENT_SCORE_ENABLED", "false") == "true"`
I've added a comment to make it clear

annotation/app/tasks/services.py
Nathicanaa: Moving this block of code in another function is pointless in my opinion because we ain't gonna reuse it anywhere else, so we'll just create a redundant function.
I believe this code should be encapsulated into this function.

annotation/alembic/versions/7cc1ed83c309_compare_agreement_scores.py
cakeinsauce: Don't we use SQLAlchemmy `Column` here? They should accept `unique=False`.

Nathicanaa: It has `unique=False` by default
__
https://github.com/epam/badgerdoc/blob/main/assets/alembic/versions/afa33cc83d57_new_fields.py#L34

khyurri (approved):

PR #141, opened 2022-12-29
refactor: use alembic migrations to setup up-to-date database for tests with pytest for 'pipelines' microservice

spiridonovfed:


pipelines/docker-compose-for-test.yaml
```diff
@@ -0,0 +1,113 @@
...
+    env_file:
+      - .env
+    environment:
+      - DB_URI=postgresql+psycopg2://postgres:postgres@db:5432/pipelines
+      - S3_CREDENTIALS_PROVIDER=minio
+      - S3_ENDPOINT=minio:9000
+      - S3_ACCESS_KEY=minioadmin
+      - S3_SECRET_KEY=minioadmin
+      - PREPROCESSING_URL=//preprocess:8000/
+      - POSTPROCESSING_URI=//postprocess:8004/
```
cakeinsauce, line 80: Are these slashes at the beginning and the end intentional?

pipelines/tests/conftest.py
```diff
@@ -1,67 +1,56 @@
...
-    create_database(testing_engine.url)
-    dbm.Base.metadata.create_all(bind=testing_engine)
+def testing_session(testing_engine, use_temp_env_var):
+    if not database_exists(testing_engine.url):
+        create_database(testing_engine.url)
+
+    try:
+        command.upgrade(alembic_cfg, "head")
+    except SQLAlchemyError as e:
+        raise SQLAlchemyError(f"Got an Exception during migrations - {e}")
```
cakeinsauce, line 45: May we move db creation related stuff to another fixture? In this case `testing_engine` I believe. Or better create a new one, like `setup_db`

pipelines/tests/test_execution.py
```diff
@@ -58,6 +58,10 @@ async def test_step_execution_with_logging(
     assert step_exec_mock.call_count == 1
 
 
+@pytest.mark.skip(
+    "Test should be fixed - it 'blinks'. "
+    "It passes when run separately, but fails when all tests are run."
+)
```
cakeinsauce, line 64: May you create an issue for this?

pipelines/docker-compose-for-test.yaml
spiridonovfed: They are.
I've copied it from existing (but not entirely properly working - that's why I don't use it) docker-compose.yaml.

A bit more about this syntax:
https://stackoverflow.com/questions/9646407/two-forward-slashes-in-a-url-src-href-attribute#:~:text=The%20%22two%20forward%20slashes%22%20are,to%20load%20the%20current%20page%22.&text=type%20of%20logic%20all%20over,through%20both%20http%20and%20https%20).

pipelines/tests/test_execution.py
spiridonovfed: For sure.
I will create an issue right after merging to main.
(They will be marked 'skip' in 'main' branch and it would be easier to reffer to them

pipelines/tests/conftest.py
spiridonovfed: agreed - moved relevant logic to 'setup_test_db' fixture

cakeinsauce (approved):

PR #142, opened 2022-12-29
fix: add check for taxonomy in get_taxonomy function

### LINKED ISSUE #140: Error when no taxonomy for primary key in get_taxonomy function
image

thinklab:
closes #140

### DIFF
--- a/taxonomy/app/taxonomy/services.py
+++ b/taxonomy/app/taxonomy/services.py
@@ -37,7 +37,7 @@ def get_taxonomy(
     tenant: str,
 ) -> Optional[Taxonomy]:
     taxonomy = session.query(Taxonomy).get(primary_key)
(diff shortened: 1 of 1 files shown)


taxonomy/app/taxonomy/services.py
```diff
@@ -37,7 +37,7 @@ def get_taxonomy(
     tenant: str,
 ) -> Optional[Taxonomy]:
     taxonomy = session.query(Taxonomy).get(primary_key)
-    if taxonomy.tenant in (tenant, None):
+    if taxonomy and taxonomy.tenant in (tenant, None):
```
Andrka, line 40: This should also be changed in line 84

spiridonovfed (approved):
Code looks good, but please test your fix and make sure that all tests are run with no fails

PR #152, opened 2023-01-03
refactor: fix processing bootstrap, several tests, change test_db setup to alembic migrations

spiridonovfed:
…p to alembic migrations


processing/Dockerfile
```diff
@@ -14,23 +14,15 @@ ENV PYTHONUNBUFFERED 1
 FROM base as build
 COPY alembic alembic
 COPY src src
-CMD alembic upgrade head && uvicorn src.main:app --host 0.0.0.0 --port 8080
+CMD alembic upgrade head && uvicorn src.main:app --host 0.0.0.0 --port 8080 --reload
```
cakeinsauce, line 17: May you clarify usage of `reload` here?

processing/tests/conftest.py
```diff
@@ -67,24 +74,43 @@ def files_data_for_pipeline(files_data):
...
-    Base.metadata.create_all(bind=engine)
-    session_local = sessionmaker(
-        autocommit=False, autoflush=False, bind=engine
-    )
-    session = session_local()
-    yield session
-    session.close()
-    Base.metadata.drop_all(bind=engine)
-    remove_db(test_db_name)
+def test_db_engine():
```
cakeinsauce, line 87: May you rename it to `testing_db_engine` or `db_test_engine` so we won't confuse it with tests

processing/tests/conftest.py
```diff
@@ -67,24 +74,43 @@ def files_data_for_pipeline(files_data):
...
+    except SQLAlchemyError as e:
+        raise SQLAlchemyError(f"Got an Exception during migrations - {e}")
+
+    yield
+
+    drop_database(test_db_engine.url)
+
+
+@pytest.fixture
+def test_db_session(test_db_engine, setup_database):
```
cakeinsauce, line 110: May you rename it the same way as testing engine?

processing/Dockerfile
spiridonovfed: I used it during development so the webapp could reload after changes in code.
I've deleted it here

processing/tests/conftest.py
spiridonovfed: agreed - renamed to db_test_engine

processing/tests/conftest.py
spiridonovfed: agreed - renamed to db_test_session

cakeinsauce (approved):

PR #159, opened 2023-01-04
refactor: use alembic migrations to setup up-to-date database for tests in 'taxonomy'

spiridonovfed:


taxonomy/app/database.py
```diff
@@ -17,6 +18,16 @@
...
 engine = create_engine(SQLALCHEMY_DATABASE_URL)
 
+# Ensure LTREE extensions is installed
+with engine.connect() as conn:
+    try:
+        conn.execute(sqlalchemy.sql.text("CREATE EXTENSION IF NOT EXISTS ltree"))
+    except sqlalchemy.exc.ProgrammingError as err_:
+        # Exctension installed, just skip error
+        if "DuplicateObject" not in str(err_):
+            raise err_
```
cakeinsauce, line 28: May we move it to some sort of function? and add it as pre-setup step for app?

cakeinsauce: For instance, if it's not done already, create setup_db function to adjust all db settings before starting the app

spiridonovfed: I've decided to move it directly to alembic migrations as long as it set our database up and it seems appropriate having this logic there

cakeinsauce (approved):
LGTM!
Squash commits before merging, please

PR #161, opened 2023-01-04
fix: add app exceptions middleware with logger

### LINKED ISSUE #87: Psycopg2 exceptions are not caught by FastAPI error handlers.
Psycopg2 is used as a backbone for sqlalchemy.
Errors from this lib are hot handled by FastAPI handler even with 'Exception' provided as a base class. 

I've met this when table was not updated properly with migration and insert caused errors.

thinklab:
fixes #87

### DIFF
--- a/annotation/app/main.py
+++ b/annotation/app/main.py
@@ -76,7 +76,7 @@ async def catch_exceptions_middleware(request: Request, call_next):
         return await call_next(request)
     except Exception as exception:  # noqa
         logger.exception(exception)
-        return Response("Internal server error", status_code=500)
+        raise exception
 
 app.middleware('http')(catch_exceptions_middleware)
 app.include_router(annotations_resources.router)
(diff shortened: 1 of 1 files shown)


annotation/app/main.py
```diff
@@ -65,6 +68,17 @@ def get_version() -> str:
...
 
+logger = app_logger.Logger
+
+
+async def catch_exceptions_middleware(request: Request, call_next):
+    try:
+        return await call_next(request)
+    except Exception as exception:  # noqa
+        logger.exception(exception)
+        return Response("Internal server error", status_code=500)
```
cakeinsauce, line 79: Shouldn't we re-raise an exception after logging? Because we may/will have a middleware to handle those exceptions on our own, at this point we simply conceal them.
```suggestion
    except Exception as exception:  # noqa
        logger.exception(exception)
        raise exception
```

cakeinsauce (approved):

PR #162, opened 2023-01-05
fix: task responce changed

sbilevich:


web/src/api/typings/tasks.ts
```diff
@@ -12,11 +12,18 @@ export type Task = {
         name: string;
         id: number;
     };
-    user_id: string;
+    user: {
+        id: string;
+        name: string;
+    };
```
Iogsotot, line 18: we have a user object with an id field and a separate user_id field, right? what is the difference between these id?

web/src/api/typings/tasks.ts
```diff
@@ -1,6 +1,6 @@
 import { ValidationType } from 'api/typings';
 
-export type Task = {
+export type ApiTask = {
```
Iogsotot, line 3: Can you clarify for me please: does taskApi mean a task that comes from the backend?

web/src/api/typings/tasks.ts
```diff
@@ -12,11 +12,18 @@ export type Task = {
...
-    user_id: string;
+    user: {
+        id: string;
+        name: string;
+    };
     is_validation: boolean;
     deadline: string;
 };
 
+export type Task = Omit<ApiTask, 'user'> & {
```
Iogsotot, line 23: What is the difference between entities Task and ApiTask?

Iogsotot (approved):

web/src/api/typings/tasks.ts
sbilevich: yes

web/src/api/typings/tasks.ts
sbilevich: before changes on backend there was user_id, now user object. I created new type for task that comes from backend, so we can take user id from it and use it on front-end without changing Task type

web/src/api/typings/tasks.ts
sbilevich: Task has user_id, ApiTask user:{id, name}

PR #189, opened 2023-01-10
refactor: setup test database with migrations

spiridonovfed:


scheduler/tests/conftest.py
```diff
@@ -1,49 +1,54 @@
...
 @pytest.fixture
-def testing_session(testing_engine):
+def testing_session(use_temp_env_var, testing_engine):
     sqlalchemy_utils.create_database(testing_engine.url)
-    models.Base.metadata.create_all(bind=testing_engine)
+
+    try:
+        command.upgrade(alembic_cfg, "head")
+    except SQLAlchemyError as e:
+        raise SQLAlchemyError(f"Got an Exception during migrations - {e}")
```
cakeinsauce, line 50: Why should we raise the same exception with slightly adjusted message and lost traceback?

spiridonovfed: By doing this I give an understanding, that this Exception is raised specifically during migrations in explicit way.
It helped me a lot during development

cakeinsauce:
```suggestion
    except SQLAlchemyError as e:
        raise SQLAlchemyError("Exception during migrations") from e
```

spiridonovfed: Agreed - applied your suggestion

cakeinsauce (approved):

PR #193, opened 2023-01-10
fix:invalidate tasks after save

sbilevich:


web/src/api/hooks/tasks.ts
```diff
@@ -280,25 +280,37 @@ export const useGetValidatedPages: QueryHookType<UseTaskForValidatedPagesParams,
...
           }
         : {};
-    return useBadgerFetch({
+    return await useBadgerFetch<TaskModel>({
         url: `${namespace}/tasks/${taskId}/finish`,
         method: 'post'
     })(JSON.stringify(body));
 };
 
+export const useSetTaskFinishedMutation: MutationHookType<TaskOptions, TaskModel> = () => {
```
Iogsotot, line 304: maybe it would be better to make this function clean

iurii-topychkanov: What do you mean by this?

iurii-topychkanov (approved):

web/src/api/hooks/tasks.ts
Iogsotot: answered in Teams

Iogsotot (approved):

PR #208, opened 2023-01-11
Feature/fe update UI library#174

Iogsotot:


web/src/components/categories/categories-selection-mode-toggle/categories-selection-mode-toggle.tsx
```diff
@@ -82,8 +82,9 @@ export const CategoriesSelectionModeToggle: React.FC<SelectionModeSelectorProps>
                         onValueChange={onChangeSelectedTool}
                         direction="horizontal"
                         cx="selection-mode-toggle"
-                        RadioInput={{ checked: false }}
-                        size="12"
+                        //TODO check if this functionality & style are needed
```
iurii-topychkanov, line 85: can be removed and if it will be needed we could get it through git

iurii-topychkanov (commented):
I've totally found 3 ts-ignores and 1 PickerInput<unknown, unknown>
Looks like it's not a big deal to fix them as a part of current PR.
I think it's needed since based on previous experience such things will be successfully forgotten and it may play a bad joke in the future.

E.g New developers will come to project after us, will update epam-ui lib one more time and ts-ignore will hide contract problems if they will be

web/src/connectors/categories-table-connector/categories-table-connector.tsx
```diff
@@ -46,7 +46,8 @@ export const CategoriesTableConnector: FC<CategoriesTableConnectorProps> = () =>
         },
         [data?.data]
     );
-
+    //TODO need to fix
+    //@ts-ignore
```
sbilevich, line 50: can we try to get rid of ts-ignors?

Iogsotot: I suppose it would be better to remove all ts-ignore in project (and maybe some of them are connected, so that's why it need to be removed together)
image

sbilevich: yes that would be great to get rid of all ts-ignors, but it should be done in a refactoring task. Here we are discussing this PR and it would be great if you will not add new ts-ignors to the project. If it's a difficult change then it could be done in a separate task.

web/src/components/task/task-sidebar/task-sidebar.tsx
```diff
@@ -408,24 +409,37 @@ const TaskSidebar: FC<TaskSidebarProps> = ({ onRedirectAfterFinish, jobSettings,
...
+                                                        renderName: (s: any) => (
+                                                            <span
+                                                                style={{
+                                                                    color: el.metadata?.color
+                                                                }}
+                                                            >
+                                                                {el.name}
+                                                            </span>
+                                                        )
+                                                    })) as unknown as RadioGroupItem<{
```
iurii-topychkanov, line 436: It's not the right way to do it, since when we set unknown first it allows as to assert any type later. So if API will change it won't show any problems here. However it's better than nothing.

iurii-topychkanov (approved):

Issue #182, opened 2023-01-09
Highlight taxon as selected when clicking on annotation with taxon label


sbilevich:
Go to annotation task
Select an annotation with taxon label
Selected taxon name should be highlighted in Data tab

![image](https://user-images.githubusercontent.com/69203629/211288239-a934df12-a36e-4812-930c-b7ee3245801c.png)

PR #234, opened 2023-01-19
fix: select related taxon item in Data tab on clicking an annotation

Iogsotot:
Fixed taxon highlighting as selected when clicking on an annotation with a taxon label.
Closes https://github.com/epam/badgerdoc/issues/182
image


web/.gitignore
```diff
@@ -1,5 +1,9 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+# local extensions & settings
```
iurii-topychkanov, line 3: I would not place local extensions & settings to global gitignore, imagine situation when each developer in team has 10 different extensions, it will become madness to maintain this file. Instead you can not place this files to stash or just not track changes

web/src/components/taxonomies/taxonomies-tree.tsx
```diff
@@ -53,23 +55,27 @@ export const TaxonomiesTree: FC<TaxonomiesTreeProps> = ({
...
 
-    const handleSelect = useCallback((selectedKeys, info) => {
-        setSelectedKeys(selectedKeys);
-        if (selectedAnnotation?.labels) {
-            const labelToChangeIdx: number = selectedAnnotation?.labels.findIndex(
-                (label: AnnotationLabel) => label.annotationId === selectedAnnotation.id
-            );
+    const handleSelect = useCallback(
+        (handleSelectedKeys, info) => {
+            selectedKeys = handleSelectedKeys;
```
iurii-topychkanov, line 60: hmm, it's look like antipattern for react apps, instead it would be better to change selectedKeys to state

web/src/components/taxonomies/taxonomies-tree.tsx
```diff
@@ -53,23 +55,27 @@ export const TaxonomiesTree: FC<TaxonomiesTreeProps> = ({
...
     );
 
-    const handleSelect = useCallback((selectedKeys, info) => {
-        setSelectedKeys(selectedKeys);
-        if (selectedAnnotation?.labels) {
-            const labelToChangeIdx: number = selectedAnnotation?.labels.findIndex(
-                (label: AnnotationLabel) => label.annotationId === selectedAnnotation.id
-            );
+    const handleSelect = useCallback(
+        (handleSelectedKeys, info) => {
```
sbilevich, line 59: let's rename handleSelectedKeys to selectedNodes for example, as it's not a function

web/src/components/taxonomies/taxonomies-tree.tsx
Iogsotot: but these changes fixed problems with multiple re-renders. Could you send me a link for this antipattern, pls? I want to read more about this

Iogsotot: it helps with state in this case. Another solution requires a deep refactoring of the project

iurii-topychkanov (approved):

sbilevich (approved):

PR #297, opened 2023-02-01
feat: enabled agreement score service

MagicTearsAsunder:


annotation/app/errors.py
```diff
@@ -156,5 +156,5 @@ def agreement_score_service_error_handler(
 ):
     return JSONResponse(
         status_code=400,
-        content={"detail": f"Agreement score service error. {exc.exc_info}"},
+        content={"detail": f"Agreement score service error. {exc}"},
     )
```
spiridonovfed, line 160: I usually do:
try:
    ...
except SomeException as exc:
    raise HTTPException(status_code=400, detail="Agreement score service error. {exc}"

But it's up to you

annotation/app/schemas/tasks.py
```diff
@@ -152,6 +152,7 @@ class AgreementScoreServiceInput(BaseModel):
     task_id: int = Field(..., example=1)
     s3_file_path: str = Field(...)
     s3_file_bucket: str = Field(...)
+    s3_tokens_path: str = Field(...)
     manifest_url: str = Field(...)
```
spiridonovfed, line 156: s3_file_path: str = Field(...)
is the same as:
s3_file_path: str

There's no point of using "= Field(...)" unless you use some additional attributes, like 'example' in line 152

annotation/app/tasks/resources.py
```diff
@@ -920,7 +920,11 @@ def finish_task(
 
             # TODO extensive coverage, annotators, manifest_url
             extensive_coverage = task.jobs.extensive_coverage
-            if extensive_coverage and extensive_coverage > 1:
+            if (
+                extensive_coverage
+                and extensive_coverage > 1
+                and AGREEMENT_SCORE_ENABLED == "true"
```
spiridonovfed, line 926: it's up to you, but this way seems more Pythonic to me:

#.env
AGREEMENT_SCORE_ENABLED=0

#code.py
if AGREEMENT_SCORE_ENABLED:
    ....

annotation/app/tasks/services.py
```diff
@@ -539,21 +540,38 @@ def evaluate_agreement_score(
...
+        .all()
+    )
+    tasks_intersection_pages.append(task)
+
     s3_file_path, s3_file_bucket = get_file_path_and_bucket(
         task.file_id, tenant, token.token
     )
+    # s3_file_path, s3_file_bucket = (
+    #     f"files/{task.file_id}/{task.file_id}.pdf", tenant
+    # )
```
spiridonovfed, line 562: looks like redundant code

spiridonovfed (changes requested):

annotation/app/errors.py
MagicTearsAsunder: This is a common error handler

annotation/app/tasks/resources.py
MagicTearsAsunder: I think this is less explicit

annotation/app/tasks/services.py
MagicTearsAsunder: fixed

annotation/app/schemas/tasks.py
MagicTearsAsunder: fixed

spiridonovfed (approved):

PR #298, opened 2023-02-02
fix: extensive coverage validation

sbilevich:


web/.gitignore
```diff
@@ -29,10 +29,7 @@ yarn-debug.log*
 yarn-error.log*
 
 # local env files
-.env.local
-.env.development.local
-.env.test.local
-.env.production.local
```
Iogsotot, line 35: I'm not sure if keeping .env in the repo is a good practice

Iogsotot (changes requested):

Iogsotot (approved):

PR #302, opened 2023-02-02
fix: change agreement metrics flow

MagicTearsAsunder:
Deleted AgreementScore table, agreement scores are stored using only AgreementMetrics table. Agreement score export from task stats retrives a json containg task_to (initial task), task_from (a task compared with) and agreement_metric. Fixed all related tests.


annotation/app/tasks/services.py
```diff
@@ -557,6 +551,9 @@ def evaluate_agreement_score(
     s3_file_path, s3_file_bucket = get_file_path_and_bucket(
         task.file_id, tenant, token.token
     )
+    # s3_file_path, s3_file_bucket = (
```
andrei-shulaev, line 554: Shall we delete it?

MagicTearsAsunder: deleted

andrei-shulaev (approved):

Issue #250, opened 2023-01-25
 Cannot configure display of path from the root to a node of taxonomy in working area


DaryaGurtovenko:
_Issue came from UAT_

**Preconditions:**
1. A hierarchical taxonomy exists.
2. An extraction job is created with category linked to the hierarchical taxonomy

**Steps to reproduce:**
1. Open annotation task.
2. Select the category associated with the taxonomy and create a label. The right panel switches to the “data” tab.
3. Choose some taxon that is not at the top level of the taxonomy (i.e., has ancestors).

**Result:**
On the assigned tag above the highlighted text the full path from the taxonomy root is not shown, and there’s no way to configure 
it in the tool
![image](https://user-images.githubusercontent.com/80756170/214523195-5ddaec22-a50c-4abe-a9b1-deb5df0e0cb1.png)

sbilevich:
Works as expected, needs to be done as a feature

MarinaVorobeva21:
Verified. 
The full path goes with dots.

MarinaVorobeva21:
Reopened as reproduced on dev1. But there should be a decision if we need to fix it as we have a full path on the sidebar now.
![2023-04-13_23h15_14](https://user-images.githubusercontent.com/98399603/231884840-6153613d-4cac-49fc-b3ff-f0a6500f2ea2.png)

PR #317, opened 2023-02-09
feat: add parents_concatenate taxon name and id endpoint

MagicTearsAsunder:
Issue: #250

### DIFF
--- a/taxonomy/app/schemas/taxon.py
+++ b/taxonomy/app/schemas/taxon.py
@@ -39,3 +39,18 @@ class TaxonResponseSchema(TaxonInputSchema):
     class Config:
         allow_population_by_field_name = True
         orm_mode = True
+
+
+class ParentsConcatenateResponseSchema(BaseModel):
+    taxon_id: str = Field(..., example="my_taxon_id")
+    taxon_name: str = Field(..., example="taxon_name")
+    parent_ids_concat: Optional[str] = Field(
+        ..., example="parent_id_1.parent_id_2"
+    )
+    parent_names_concat: Optional[str] = Field(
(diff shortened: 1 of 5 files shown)


taxonomy/app/schemas/taxon.py
```diff
@@ -39,3 +39,18 @@ class TaxonResponseSchema(TaxonInputSchema):
...
+    taxon_name: str = Field(..., example="taxon_name")
+    parent_ids_concat: Optional[str] = Field(
+        ..., example="parent_id_1.parent_id_2"
+    )
+    parent_names_concat: Optional[str] = Field(
+        ..., example="parent_name_1.parent_name_2"
+    )
+
+    class Config:
+        allow_population_by_field_name = True
```
spiridonovfed, line 55: You don't use aliases for model fields. So there's no point of using allow_population_by_field_name = True

taxonomy/app/schemas/taxon.py
```diff
@@ -39,3 +39,18 @@ class TaxonResponseSchema(TaxonInputSchema):
...
+        orm_mode = True
```
spiridonovfed, line 56: Do you this model with ORM?
If not - there's no point of using this attribute

taxonomy/app/schemas/taxon.py
MagicTearsAsunder: agree, removed both

taxonomy/app/schemas/taxon.py
MagicTearsAsunder: agree, removed both

spiridonovfed (approved):

PR #325, opened 2023-02-10
fix: delete build:info

sbilevich:


Iogsotot (approved):

web/package.json
```diff
@@ -55,7 +55,6 @@
         "lint:js": "eslint . --ext .js,.ts,.tsx",
         "lint:types": "tsc -p tsconfig.json --noEmit",
         "hooks:i": "npx simple-git-hooks",
-        "build:info": "node ./src/setupEnvInfo.js",
```
Iogsotot, line 58: It seems to me that it is worth removing all references to the deleted script
image

Iogsotot: and the file `setupEnvInfo.js` too

PR #343, opened 2023-02-15
add taxonomies and document_labels support to_export to label studio

spiridonovfed:
…format

### DIFF
--- a/convert/src/label_studio_to_badgerdoc/badgerdoc_to_label_studio_use_case.py
+++ b/convert/src/label_studio_to_badgerdoc/badgerdoc_to_label_studio_use_case.py
@@ -2,6 +2,9 @@ import tempfile
 from pathlib import Path
 from typing import NamedTuple
 
+from botocore.client import BaseClient
+from tenant_dependency import TenantData
+
 from src.label_studio_to_badgerdoc.badgerdoc_format.annotation_converter_practic import (
     AnnotationConverterToTheory,
 )
@@ -24,12 +27,21 @@ class BadgerdocData(NamedTuple):
 
 
 class BDToLabelStudioConvertUseCase:
+    labelstudio_format = LabelStudioFormat()
+
     def __init__(
         self,
-        s3_client,
+        s3_client: BaseClient,
+        current_tenant: str,
+        token_data: TenantData,
     ) -> None:
         self.s3_client = s3_client
-        self.labelstudio_format = LabelStudioFormat()
+        self.current_tenant = current_tenant
+        self.token_data = token_data
+        self.request_headers = {
+            "X-Current-Tenant": self.current_tenant,
+            "Authorization": f"Bearer {self.token_data.token}",
+        }
 
     def execute(
         self,
@@ -46,7 +58,10 @@ class BDToLabelStudioConvertUseCase:
             s3_input_tokens, s3_input_annotations, s3_input_manifest
         )
         self.labelstudio_format.from_badgerdoc(
-            badgerdoc_page, badgerdoc_annotations, badgerdoc_manifest
+            badgerdoc_page,
+            badgerdoc_annotations,
+            badgerdoc_manifest,
+            self.request_headers,
         )
         self.upload_labelstudio_to_s3(s3_output_annotation)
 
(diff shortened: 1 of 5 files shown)


convert/src/label_studio_to_badgerdoc/badgerdoc_to_label_studio_use_case.py
```diff
@@ -24,12 +27,21 @@ class BadgerdocData(NamedTuple):
...
+        current_tenant: str,
+        token_data: TenantData,
     ) -> None:
         self.s3_client = s3_client
-        self.labelstudio_format = LabelStudioFormat()
+        self.current_tenant = current_tenant
+        self.token_data = token_data
+        self.request_headers = {
+            "X-Current-Tenant": self.current_tenant,
+            "Authorization": f"Bearer {self.token_data.token}",
```
MagicTearsAsunder, line 43: I believe **token data** should contain token type and should not be hardcoded as "Bearer". Please, doublecheck this.

convert/src/label_studio_to_badgerdoc/labelstudio_format/label_studio_format.py
```diff
@@ -68,6 +108,17 @@ def convert_annotation_from_bd(
                     labels=[obj.category],
                 ),
             )
+            # taxon_value = [taxon_attr['value'] for taxon_attr in obj.data.dataAttributes]
+            # item.value.taxons = [taxon_value[0] if taxon_value else None for label in item.value.labels]
```
MagicTearsAsunder, line 112: Redundant?

convert/src/label_studio_to_badgerdoc/labelstudio_format/label_studio_format.py
```diff
@@ -68,6 +108,17 @@ def convert_annotation_from_bd(
...
+            if obj.data.dataAttributes:
+                item.value.taxons = [
+                    obj.data.dataAttributes[0]["value"],
+                ]
+            else:
+                item.value.taxons = [
+                    None,
+                ]
```
MagicTearsAsunder, line 120: List with single element [None] is weird (also wrong linter formatting due to redundant comma).
You can use your "safe_list_get_item" function without `if/else` here like:
```
item.value.taxons = [
    safe_list_get_item(my_list=obj.data.dataAttributes, index=0, default={}).get("value")
]
```

serereg (approved):

convert/src/label_studio_to_badgerdoc/labelstudio_format/label_studio_format.py
```diff
@@ -105,3 +156,139 @@ def form_link_direction(self, link: AnnotationLink) -> str:
...
+        except requests.exceptions.RequestException as e:
+            LOGGER.exception("Failed request to 'taxonomy' to get taxons_used")
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="Failed request to 'taxonomy' to get taxons_used",
+            ) from e
+        response_content = request_to_get_taxons_used.json()
+        LOGGER.debug(
+            "Got this response from taxonomy service: %s", response_content
+        )
```
MagicTearsAsunder, line 287: Exact request code repeats 3 times. Probably use a common function?

MagicTearsAsunder (changes requested):

convert/src/label_studio_to_badgerdoc/labelstudio_format/label_studio_format.py
spiridonovfed: Yes, redundant.
Deleted now

convert/src/label_studio_to_badgerdoc/labelstudio_format/label_studio_format.py
spiridonovfed: I would prefer to keep it this way.
Right now shows the established format of "taxons" field in label studio format files

convert/src/label_studio_to_badgerdoc/labelstudio_format/label_studio_format.py
spiridonovfed: In this file we have 1 GET and 2 POST requests and would prefer to keep it this way now.
These functions are large now because we create url variable, request body and relevant custom logging messages.
On the one hand - there are not a lot of variations and these pieces of code seem to be very close, but on the other hand - this is very easy in debugging and understanding where the problem is. I think this functionality will need a lot of debugging in near future

MagicTearsAsunder (approved):

PR #355, opened 2023-02-20
fix: In 'Files' tab of job the pagination counter does't work

### LINKED ISSUE #313: In 'Files' tab of job the left arrow is not disabled on the last page of file list
_Minor_
The right arrow also is enabled even for job with one file only, but I described the case for several pages below. 

**Preconditions:**
Extraction job with more than 15 files is created.

**Steps to reproduce:**
1. Open the job, 'Files' tab.
2. Chose with 'Show on page' drop-down amount of file when not all of them are displayed on one page ('15' for 16 files). 
3. Click on pagination arrow until the last page.

**Result:**
It's possible to click on right arrow after the last page, the error occurs:
image
In the same time the left arrow is correctly disabled on the first page of list.

**Expected result:**
The right pagination arrow should be disabled on the last page of list.

Aleksei-Egorenko-EPAM:
Closes #313


Iogsotot (approved):

web/src/connectors/job-detail-view-connector/job-detail-view-connector.tsx
```diff
@@ -99,6 +100,12 @@ export const JobConnector: React.FC<JobDetailViewProps> = ({
 
     const { data: job, refetch: refetchJob } = useJobById({ jobId }, { refetchInterval: 10000 });
 
+    useEffect(() => {
+        if (job?.files.length !== undefined) {
```
AnastasiiaPlyako, line 104: Minor issue: is it possible to write just `!job?.files.length`?

web/src/shared/hooks/page-table-hook.ts
```diff
@@ -16,7 +16,7 @@ export const usePageTable = <T, TFilter = TableFilters<T>>(item: keyof T) => {
         field: item,
         direction: SortingDirection.DESC
     });
-    const [totalCount, onTotalCountChange] = useState<number>(1);
+    const [totalCount, _onTotalCountChange] = useState<number>(1);
```
Iogsotot, line 19: Mb it will be more appropriate to use this naming: `setOnTotalCountChange`

PR #364, opened 2023-02-22
fix: validation

sbilevich:


web/src/connectors/task-annotator-connector/use-validation.tsx
```diff
@@ -124,42 +123,39 @@ export const useValidation = ({
...
         setAnnotationSaved(false);
-    }, [invalidPages, validPages, currentPage]);
+    }, [invalidPages, validPages, currentPage, notProcessedPages]);
 
     const onInvalidClick = useCallback(() => {
         if (validPages.includes(currentPage)) {
             const newValidPages = validPages.filter((page) => page !== currentPage);
             setValidPages(newValidPages);
         }
+        if (notProcessedPages.includes(currentPage)) {
```
AnastasiiaPlyako, line 152: Here we have duplicated code. Can we move it to a separate function?

AnastasiiaPlyako (approved):

PR #371, opened 2023-02-24
feat/add_next_and_previous_tasks_endpoint

spiridonovfed:


annotation/app/tasks/services.py
```diff
@@ -651,3 +652,29 @@ def save_agreement_metrics(
...
+) -> int:
+    list_of_task_ids = [task.id for task in tasks_list]
+    return list_of_task_ids.index(task_id)
+
+
+def get_previous_task(
+    tasks_list: List[ManualAnnotationTaskSchema], current_index: int
+) -> Optional[ManualAnnotationTaskSchema]:
+    previous_task_index = current_index - 1
+    if previous_task_index < 0:
```
khyurri, line 668: Let's simplify: 
If we order by id, so previous task is:
`select .. where id < current_id order by id desc limit 1`
and next is
`select ... where id > current_id order by id asc limit 1`

Andrka (commented):
Could you add some simple tests?

annotation/app/tasks/services.py
spiridonovfed: Rewritten in a suggested way

spiridonovfed:
> Could you add some simple tests?

Sure. Tests are added

Andrka (approved):

Issue #292, opened 2023-01-31
Validated version with 2 annotations instead of 3 should be available to save and to finish with warning


DaryaGurtovenko:
_Came from UAT_
During the course of the project, the requirement can be changed for business reasons, and the number of annotations is reduced to 2.

**Preconditions:**
1. Annotation job with 3 annotators is created

**Steps to reproduce:**
1. Try to finish the job with 2 annotators only.

**Expected result:**
Validated version with 2 annotations instead of 3 should be available to save and to finish with warning.

khyurri:
User, who can manage job, can remove annotator from this job and "redestribute" not finished tasks between other annotators.

@DaryaGurtovenko please, verify it's working

DaryaGurtovenko:
Admin is able to re-distribute existing job between another set (via 'Edit') of annotators, so this workaround can be used.

sbilevich:
#304 #331 should be done to fix an issue

Iogsotot:
need changes on backend side: "extensive_coverage" field returns wrong
<img width="214" alt="image" src="https://user-images.githubusercontent.com/50149163/221202695-2225c19e-5620-4f54-a137-df163b8782d5.png">
<img width="220" alt="image" src="https://user-images.githubusercontent.com/50149163/221202841-191714b2-3ba5-4b5e-89cd-5d1ca2425c7c.png">
<img width="203" alt="image" src="https://user-images.githubusercontent.com/50149163/221203064-c0fc2ab6-5adf-48d3-83f7-10e0f7ad3ee8.png">
@andrei-shulaev

zefiris87:
@andrei-shulaev 

If User try to save edited job, the following error appears:

"Failed request to the Annotation Manager: 500, message='Internal Server Error', url=URL('http://annotation/jobs/1962')"

And there is 422 response for the job request (as far as I understand)

![image](https://user-images.githubusercontent.com/124054458/223441166-31f4621b-aadd-4b9e-b45c-78f728e28519.png)

khyurri:
Will be closed during: https://github.com/epam/badgerdoc/issues/425

zefiris87:
Can't complete validated version. The following popup appears, but "Confirm validation" doesn’t work.

Job is http://dev1.badgerdoc.com/annotation/tasks/3971

![image](https://user-images.githubusercontent.com/124054458/228529164-63204ee0-868d-4f00-a688-a5a2c078fda2.png)

[...]

PR #374, opened 2023-02-24
fix: edit job#292

Iogsotot:
Fixes the frontend part of this issue: https://github.com/epam/badgerdoc/issues/292

### DIFF
--- a/web/src/connectors/edit-job-connector/edit-job-connector.tsx
+++ b/web/src/connectors/edit-job-connector/edit-job-connector.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactElement, useCallback, useContext, useMemo, useEffect } from 'react';
+import React, { FC, ReactElement, useCallback, useContext, useMemo } from 'react';
 import EditJobSettings from 'components/job/edit-job-settings/edit-job-settings';
 import { usePipelines } from 'api/hooks/pipelines';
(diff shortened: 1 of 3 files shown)


web/src/connectors/edit-job-connector/edit-job-connector.tsx
```diff
@@ -146,63 +148,73 @@ const EditJobConnector: FC<EditJobConnectorProps> = ({
...
+                pipeline
+            } = values;
+
+            if (jobType === 'NoExtraction') {
                 onRedirectAfterFinish?.();
                 return;
             }
 
+            const deadline = jobDeadline ? new Date(jobDeadline).toISOString() : undefined;
+            const annotators = validationType === 'cross' ? annotators_validators : jobAnnotator;
```
AnastasiiaPlyako, line 175: Maybe create an enum for validationType and not use magic const? I know that this approach uses in the whole project, but we can start to do on the right way)

Iogsotot: Yes, this is a good idea for the next pool of the Refazing Rebellion. Thanks!

Iogsotot: By the way, I just checked this - as far as I can see, we use typing for these constants, so it's definitely not a "magic constant"
image
image
image

Aleksei-Egorenko-EPAM (approved):

PR #382, opened 2023-02-27
fix: switch txt convert process to proper gotenberg url

thinklab:


assets/.env
```diff
@@ -31,7 +31,7 @@ GOTENBERG=gotenberg:3000
 GOTENBERG_LIBRE_OFFICE_ENDPOINT="http://${GOTENBERG}/forms/libreoffice/convert"
 GOTENBERG_CHROMIUM_ENDPOINT="http://${GOTENBERG}/forms/chromium/convert/html"
 
-GOTENBERG_FORMATS=[".docx",".doc",".bib",".xml",".csv",".fodt",".ltx",".odt",".ott",".pdb",".psw",".rtf",".sdw",".stw",".sxw",".uot",".vor",".wps",".epub",".emf",".fodg",".met",".odd",".otg",".std",".svg",".svm",".swf",".sxd",".sxw",".tiff",".xhtml",".xpm",".fodp",".potm",".pot",".pptx",".pps",".ppt",".pwp",".sda",".sdd",".sti",".sxi",".uop",".wmf",".odp"]
+GOTENBERG_FORMATS=[".txt",".docx",".doc",".bib",".xml",".csv",".fodt",".ltx",".odt",".ott",".pdb",".psw",".rtf",".sdw",".stw",".sxw",".uot",".vor",".wps",".epub",".emf",".fodg",".met",".odd",".otg",".std",".svg",".svm",".swf",".sxd",".sxw",".tiff",".xhtml",".xpm",".fodp",".potm",".pot",".pptx",".pps",".ppt",".pwp",".sda",".sdd",".sti",".sxi",".uop",".wmf",".odp"]
```
serereg, line 34: Are you sure, txt should be here?

serereg (commented):
Please look at the comment about txt in GOTENBERG_FORMATS

serereg (commented):
Separating the fix from the feature about GOTENBERG_FORMATS would be ideal. Additionally, adding support for true PDF files within that feature would be beneficial.

models/pyproject.toml
```diff
@@ -36,8 +36,8 @@ SQLAlchemy-Utils = "0.39.0"
 [tool.isort]
 profile = "black"
 
-[build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+#[build-system]
+#requires = ["poetry-core>=1.0.0"]
+#build-backend = "poetry.core.masonry.api"
```
spiridonovfed, line 41: comments out are probably redundant

assets/src/utils/common_utils.py
```diff
@@ -248,21 +248,25 @@ def convert_to_jpg(self) -> bytes:
...
-        input_text_path = f"files/{self.new_file.id}/{self.new_file.id}.txt"
+        pdf_from_txt = self.convert_to_pdf()
         output_pdf_path = f"files/{self.new_file.id}/{self.new_file.id}.pdf"
+        file_ = BytesIO(pdf_from_txt)
+        self.minio_client.put_object(
+            bucket_name=self.bucket_storage,
+            object_name=output_pdf_path,
+            data=file_,
+            length=len(pdf_from_txt),
+        )
```
spiridonovfed, line 259: Are you sure that putting the object shoudn't be done with assets?
If you put it with minio_client directly, assets will not know about it. It means, that UI will not know about it too and will not show it

spiridonovfed (changes requested):

assets/src/utils/common_utils.py
thinklab: All this preparations is in is_blank_is_created() function. There is a creation of blank record in database and new path for files.

spiridonovfed (approved):

Issue #264, opened 2023-01-25
Characters of text values in exported annotations from S3 are surrounded by spaces


DaryaGurtovenko:
Issue came from UAT

**Steps to reproduce:**
1. Download from S3 corresponding annotations JSON file and open it in a text editor.
2. In the "objs" list, find an object with the "text" field. 
 
**Result:** 
A space character is inserted between every neighboring characters of the original document text.
![image](https://user-images.githubusercontent.com/80756170/214530530-9817297d-a2dd-413e-a6ac-00056ad1f04a.png)

sbilevich:
text creates on FE

SunnyCapt:
Possible reasons for the bug:

1. An error was made on the backend when creating json - ✅ (checked, it's not)
2. An error was made on the backend when receiving json
3. Mistake made on the frontend

khyurri:
Also PR regarding new lines: https://github.com/epam/badgerdoc/pull/217

SunnyCapt:
@Iogsotot 

on frontend you can just check token.is_whitespace field to filter out the characters and concatenate them.

serereg:
Server: There must be an "after" field for the token, which must contain all the characters that come after the /s (whitespace) type token.
UI: When collecting the text field when creating a GeomObject, the front-end must, if there are more tokens after the current token in the GeomObject, then after the token you need to insert what is in the "after" field of this token.

Example: in the GeomObject, the test field should contain the text "text a". The file contained the text "this is text and not text". Tokens: text "after"=" "and "after"=" "
There is an edge case: when there are spaces at the beginning of the file. Then we add "previous"="         " to the first token. The "previous" field is not used on the UI, but is needed to assemble the text back when exporting to labelstudio format.

MarinaVorobeva21:
Cannot see the Request for a big document with several pages (e.g. [long.txt](https://github.com/epam/badgerdoc/files/10910055/long.txt)).
![image](https://user-images.githubusercontent.com/98399603/223443073-95d3ade6-a22d-45c3-9960-21ab3e3b9952.png)

serereg:
```
[...]

PR #399, opened 2023-03-02
#264 Fix: implement previous after in token

AnastasiiaPlyako:


web/src/connectors/task-annotator-connector/task-annotator-utils.ts
```diff
@@ -96,6 +96,13 @@ const addTableValues = (ann: Annotation): TableApi => {
     };
 };
 
+const createText = (token: PageInfoObjs | PageToken) => {
+    const previousSymbol = token?.previous || '';
+    const afterSymbol = token?.after || '';
+    const text = token?.text || '';
+    return previousSymbol + text + afterSymbol;
```
sbilevich, line 103: I'm not sure we need previous symbols in this attr as it's some spaces for paragraphs etc. I think afterSymbols should be enough here

sbilevich (approved):

Aleksei-Egorenko-EPAM (approved):

web/src/connectors/task-annotator-connector/task-annotator-utils.ts
AnastasiiaPlyako: It can be different symbols in the previous field. I`m afraid of the risks of throwing this functionality

sbilevich: ok, let's leave it as it is

Iogsotot (approved):

Issue #194, opened 2023-01-10
Add common linters to all microservices CI and pre-commit


khyurri:
**Run in GH Actions**

Back-end
- [ ] Black (check is formatted)
- [ ] Isort (check is sorted)
- [ ] Pylint (strict mode)
- [ ] MyPy (strict mode)
- [ ] Remove Flake8
- [ ] Bandit

Front-end
- [ ] ESLint (add to CI part)

khyurri:
Task moved here: https://github.com/epam/badgerdoc/issues/817

PR #400, opened 2023-03-03
ci: add convert github action

thinklab:


.github/workflows/convert.yml
```diff
@@ -0,0 +1,56 @@
...
+on:
+  pull_request:
+    paths:
+      - convert/**
+      - .github/workflows/convert.yml
+jobs:
+  convert-pre-commit-actions:
+    strategy:
+      matrix:
+        python-version: [ "3.8.15" ]
```
spiridonovfed, line 12: Base image uses 3.8.10.
I believe your environment for tests and linters should be consistent with that.

.github/workflows/convert.yml
```diff
@@ -0,0 +1,56 @@
...
+          cache: "pip"
+      - name: Run isort
+        uses: isort/isort-action@v1.1.0
+        with:
+          sort-paths: convert/*
+          configuration:
+            --check-only
+            --profile=black
+            --line-length=79
+            --skip=tests
```
spiridonovfed, line 28: I would also add --skip = alembic, because sometimes automatically created alembic revisions are not consistent with black linter.
Please see, let's say, .github/workflows/jobs.yaml

.github/workflows/convert.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: Black
+        uses: psf/black@stable
+        with:
+          options: "--line-length=79"
+      - run: pip install flake8
+      - run: flake8 --extend-ignore=E203 convert
```
spiridonovfed, line 34: 'alembic' and 'tests' should be escaped. See relevant commands in .github/workflows/jobs.yaml

.github/workflows/convert.yml
```diff
@@ -0,0 +1,56 @@
...
+          python-version: ${{ matrix.python-version }}
+          cache: "poetry"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          cd convert
+          poetry install --all-extras
+          poetry add --editable ../lib/filter_lib
+          poetry add --editable ../lib/tenants
+          poetry run pytest
```
spiridonovfed, line 56: Please see this issue:
https://github.com/epam/badgerdoc/issues/194

I believe it's worth checking with our teamlead about shoud you implement this issue here or not.

spiridonovfed (changes requested):

serereg:
Duplicated by https://github.com/epam/badgerdoc/pull/567

Issue #403, opened 2023-03-03
Ability to annotate the first character in the document


Iogsotot:
https://user-images.githubusercontent.com/50149163/222711659-2e81dbdf-9b40-452c-91f8-5860d0d63e34.mp4

PR #409, opened 2023-03-07
fix: ability to annotate first character in  document#403

Iogsotot:
- [x] closes https://github.com/epam/badgerdoc/issues/403
- [x] fixed naming for better readability
- [x] fixed a bug with rendering a list of components without a key in React
- [x] fixed typo


web/src/shared/components/annotator/annotator.tsx
```diff
@@ -180,7 +180,7 @@ export const Annotator: FC<AnnotatorProps> = ({
 
     const {
         coords: selectionCoords,
-        isStarted,
+        isStarted: isSelectionStarted,
```
Aleksei-Egorenko-EPAM, line 183: Could you please move the declaration of the variable above its usage?

sbilevich (approved):

sbilevich (approved):

PR #412, opened 2023-03-09
feat: add annotator's username to get_job response

spiridonovfed:


users/chart/templates/deployment.yaml
```diff
@@ -93,6 +93,16 @@ spec:
...
+#        - name: KEYCLOAK_ROLE_ClientViewsAllUsers_ID
+#          valueFrom:
+#            secretKeyRef:
+#              name: users
+#              key: KEYCLOAK_ROLE_ClientViewsAllUsers_ID
+#        - name: KEYCLOAK_ROLE_ClientViewsAllUsers_SECRET
+#          valueFrom:
+#            secretKeyRef:
+#              name: users
+#              key: KEYCLOAK_ROLE_ClientViewsAllUsers_SECRET
```
serereg, line 105: Should it be commented? Or removed at all?

jobs/.env
```diff
@@ -2,5 +2,6 @@ PIPELINES_URI = "http://127.0.0.1:8001"
 ASSETS_URI = "http://127.0.0.1:8000"
 ANNOTATION_MICROSERVICE_URI = "http://127.0.0.1:8002"
 TAXONOMY_URI = "http://127.0.0.1:8002"
+USERS_HOST = "http://dev2.badgerdoc.com/api/v1/users"
```
khyurri, line 5: Is it good idea to hardcode external URL here? Why not to use internal, like `TAXONOMY_URI`?

jobs/jobs/utils.py
```diff
@@ -583,3 +586,54 @@ def get_taxonomy_links(
         )
         for category_link in categories_links
     ]
+
+
+@alru_cache(maxsize=256)
```
khyurri, line 591: What is username will change? 
As I can see, this function doesn't implement any cache invalidation handlers

users/chart/templates/deployment.yaml
```diff
@@ -93,6 +93,16 @@ spec:
             secretKeyRef:
               name: users
               key: ADMIN_CLIENT_SECRET_DEV1
+#        - name: KEYCLOAK_ROLE_ClientViewsAllUsers_ID
```
khyurri, line 96: Why these fields was commented?

users/.env
```diff
@@ -19,6 +19,9 @@ KEYCLOAK_ENDPOINT = "http://dev2.badgerdoc.com/"
 KEYCLOAK_REALM = "master"
 KEYCLOAK_ROLE_ADMIN = "admin"
 
+KEYCLOAK_ROLE_ClientViewsAllUsers_ID=ClientViewsAllUsers
+KEYCLOAK_ROLE_ClientViewsAllUsers_SECRET=5ce0b47a-fc6a-4721-aba0-0b4d00134f5e
```
khyurri, line 23: Is it good idea to store SECRET in `.env` file?

jobs/.env
spiridonovfed: Anyway, this value is rewritten in deployment.yaml chart.
I am deleting this line in order to avoid any confusion

users/chart/templates/deployment.yaml
spiridonovfed: It is commented up to the point our DevOps engineer will put those values in 'secret' on dev environments.
It will be uncommented and used after that

jobs/jobs/utils.py
spiridonovfed: Current solution does not have any scenario in case username changes and doesn't have any cache invalidations.
If we imply that username could be changed and we have to handle it - it means no caching possible.

users/chart/templates/deployment.yaml
spiridonovfed: It is commented up to the point our DevOps engineer will put those values in 'secret' on dev environments.
It will be uncommented and used after that

users/.env
spiridonovfed: It is not and I am using it only in testing purposes.
Waiting for our DevOps engineer putting those values in 'secret' on dev environments.
It will be deleted after that

users/src/main.py
```diff
@@ -148,6 +149,29 @@ async def get_user(
...
+    token: TenantData = Depends(tenant),
+    current_tenant: Optional[str] = Header(None, alias="X-Current-Tenant"),
+) -> kc_schemas.User:
+    """Get username by its user_id with any valid token."""
+    auth_data = OAuthRequest(
+        grant_type="client_credentials",
+        client_id=conf.KEYCLOAK_ROLE_ClientViewsAllUsers_ID,
+        client_secret=conf.KEYCLOAK_ROLE_ClientViewsAllUsers_SECRET,
+    )
+    token_data = await kc_query.get_token_v2(realm, auth_data)
```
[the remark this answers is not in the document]
spiridonovfed: Does this mean to turn on @alru_cache for the coroutine kc_query.get_token_v2?

jobs/jobs/utils.py
spiridonovfed: Also, 'users' API does not support changing any attributes of any user, including username.
But it is still possible for an admin in Keycloak directly.
I am not sure whether we should keep in mind this possibility

spiridonovfed: Deleted

khyurri (approved):

Issue #415, opened 2023-03-10
Attempt to finish completed task (extensive coverage)


Andrka:
![2023-03-10_12-35](https://user-images.githubusercontent.com/14347940/224269115-9131ec8a-33d0-4b13-8168-989829972084.png)

### Preconditions:

Annotation job with extensive coverage is finished.

### Steps to reproduce:

Go to validation task (sometimes it occurs from the second time).

### Expected result:

There is no error.

zefiris87:
@Iogsotot 

I'm not sure if it's related issue, but it looks very similar

Logged in as annotator3

1. Work with task: http://dev1.badgerdoc.com/annotation/tasks/3723
2. Try to finish annotation

Case failed. There is an error:

<img width="712" alt="image" src="https://user-images.githubusercontent.com/124054458/226862128-d362ba80-feb8-4632-a9a0-5cfdc345248a.png">

Iogsotot:
I was only able to reproduce this for `task id` 3723
`http://dev1.badgerdoc.com/api/v1/annotation/docs#/Manual%20Annotation%20Tasks/finish_task_tasks__task_id__finish_post`
Login as Annotator 3:
`http://dev1.badgerdoc.com/api/v1/annotation/tasks/3723/finish`

```
{
  "detail": "Internal server error. 0"
}
curl -X 'POST' \
  'http://dev1.badgerdoc.com/api/v1/annotation/tasks/3723/finish' \
  -H 'accept: application/json' \
  -H 'X-Current-Tenant: test' \
[...]

PR #419, opened 2023-03-13
fix: attempt to finish completed task extensive coverage#415

Iogsotot:
closes https://github.com/epam/badgerdoc/issues/415

### DIFF
--- a/web/src/shared/components/status/status.tsx
+++ b/web/src/shared/components/status/status.tsx
@@ -1,10 +1,11 @@
 import React, { FC } from 'react';
+import { JobStatus } from 'api/typings/jobs';
 import { Text, Tooltip } from '@epam/loveship';
 import styles from './status.module.scss';
 
 type StatusProps = {
     color: string;
-    statusTitle: string;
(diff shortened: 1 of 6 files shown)


web/src/shared/components/status/status.tsx
```diff
@@ -1,10 +1,11 @@
 import React, { FC } from 'react';
+import { JobStatus } from 'api/typings/jobs';
 import { Text, Tooltip } from '@epam/loveship';
 import styles from './status.module.scss';
 
 type StatusProps = {
     color: string;
-    statusTitle: string;
+    statusTitle: JobStatus | string;
```
Aleksei-Egorenko-EPAM, line 8: Why we still need string?

Iogsotot: because the project has a lot of assumptions in TS, so if you remove the typing as a string at this point, then there will be a wave-like generation of errors throughout the project.

Aleksei-Egorenko-EPAM (approved):

Issue #18, opened 2022-11-11
Add Readme.md with list of all contributes


MikhailFokanov:

MysterionRise:
Should be closed via #423. @khyurri could you close it?

khyurri:
@MikhailFokanov 
Can you notify our internal contributors, that they can add themselves into Readme.md via PR?

PR #423, opened 2023-03-14
feat: add contributors list and script for Markdown generation

MysterionRise:
Fixing #18


khyurri (changes requested):
Thank you for this PR, may I ask you to move `generate_contributors.py` into folder like `scripts/` or similar. We're keeping root as clear as possible

MysterionRise:
For sure, will do that

khyurri (approved):

PR #424, opened 2023-03-14
feat: adding GH actions for multiple docker hub pushes

MysterionRise:
Trying to solve #145 
Expecting that tags should work for you as version in Docker hub

Would also require

- [ ] User names and password for the account to be available in secrets 

Happy to provide updates to logic if needed


khyurri:
@MysterionRise 
Currently we're looking for solution to host our release images in public docker registry

Let's keep this PR open until we'll make decision: docker hub or github (maybe something else)

We'll be glad to hear your opinion

PR #435, opened 2023-03-16
Replace old Job type with new extended type

nanbratan:


web/src/api/typings/jobs.ts
```diff
@@ -45,6 +45,8 @@ export type Job = {
     // TODO: check pipeline property
 };
 
+export type TJobWithExtendedAnnotators = Omit<Job, 'annotators'> & { annotators: TUserShort[] };
```
Aleksei-Egorenko-EPAM, line 48: We have already changed the description of the Job type, taking into account the received data, so these changes are redundant.

nanbratan: Done, rolled back TJobWithExtendedAnnotators related changes

Iogsotot (approved):

PR #448, opened 2023-03-18
WIP: local run

khyurri:
This PR adds docker-compose to setup local BadgerDoc instance


README.md
```diff
@@ -19,31 +21,65 @@ For now, BadgerDoc is working with vectorized and scanned documents, but it has
...
+
 ```
-&& black src --check && mypy src && pylint src \
+make clean
 ```
-5. Build the image, specifying the tag that is used in other Docker images as a "base" image:
+
+## Building microservices
+
+Easies way to build microservices is to run `make build_all` command, right after that,
```
cakeinsauce, line 41:
```suggestion
Easiest way to build microservices is to run `make build_all` command, right after that, 
```

lib/tenants/src/dependency.py
```diff
@@ -87,7 +97,8 @@ def decode_hs256(self, token: str) -> Dict[str, Any]:
                 status_code=status.HTTP_401_UNAUTHORIZED,
                 detail="Token is expired!",
             )
-        except (jwt.PyJWTError, jwt.exceptions.DecodeError):
+        except (jwt.PyJWTError, jwt.exceptions.DecodeError) as err:
+            logger.exception(err)
```
cakeinsauce, line 101: Exception logger will contain whole error traceback with messages, basically duplicating. We should probably specify own message here.
```suggestion
            logger.exception("Cannot decode invalid token")
```

lib/tenants/src/dependency.py
```diff
@@ -105,7 +116,8 @@ def decode_rs256(self, token: str) -> Dict[str, Any]:
...
```
cakeinsauce, line 120: Same as previous comment
```suggestion
            logger.exception("Cannot decode invalid rs256 token")
```

users/users/keycloak/schemas.py
```diff
@@ -326,20 +331,31 @@ class RefreshTokenRequest(OAuthRequest):
...
+        token = await admin_access_token.token()
+        # .. and retry request
+
+    TODO: think about to add refresh by refresh token
+    """
+
+    def __init__(self) -> None:
+        self.__token = None
+
+    async def token(self):
```
cakeinsauce, line 353: Could you specify return typing?

users/users/keycloak/schemas.py
```diff
@@ -326,20 +331,31 @@ class RefreshTokenRequest(OAuthRequest):
...
+        if self.__token is None:
+            await self.request_new_token()
+        return self.__token
```
cakeinsauce, line 356: If empty token (string) is not suitable as well, then we could
```suggestion
        return self.__token or await self.request_new_token()
```

users/users/keycloak/utils.py
```diff
@@ -17,20 +18,20 @@ def create_filters(users: Users) -> Dict[str, Union[str, None]]:
...
+        except aiohttp.ClientResponseError as err:
+            Logger.exception(err)
+            if err.status in {401, 403}:
+                Logger.warning("Possible access Token is expired")
+                Logger.info("Retrying...")
                 await access_token.request_new_token()
-                await asyncio.sleep(1)
             else:
-                raise e
+                raise err
```
cakeinsauce, line 37: `raise err` will remove traceback. We can use empty raise to re-raise the same exception with the same traceback.
```suggestion
                raise
```

users/users/keycloak/utils.py
```diff
@@ -17,20 +18,20 @@ def create_filters(users: Users) -> Dict[str, Union[str, None]]:
...
-            return await query.get_user(realm, access_token.token, user_id)
-        except aiohttp.ClientResponseError as e:
-            if e.status in {401, 403}:
-                Logger.info(
-                    "Access Token for ClientViewsAllUsers_Role is expired. "
-                    "Creating new one"
-                )
+            return await query.get_user(realm, token, user_id)
+        except aiohttp.ClientResponseError as err:
+            Logger.exception(err)
```
cakeinsauce, line 31: Same as previous comment
```suggestion
            logger.exception("Couldn't fetch username")
```

users/users/keycloak/utils.py
```diff
@@ -17,20 +18,20 @@ def create_filters(users: Users) -> Dict[str, Union[str, None]]:
...
+            if err.status in {401, 403}:
+                Logger.warning("Possible access Token is expired")
+                Logger.info("Retrying...")
```
cakeinsauce, line 34:
```suggestion
            if err.status in {401, 403}:
                Logger.warning("Possible access Token is expired")
            Logger.info("Retrying...")
```

lib/tenants/src/dependency.py
khyurri: > Exception logger will contain whole error traceback with messages
Yeah, that what `logger.exception` does :) 
> basically duplicating
Check next lines, it raises new `HTTPException` without `from`. So, there is no any chance to get stack trace down stream. Therefore I added explicit logging here

users/users/keycloak/utils.py
khyurri: I deleted this piece of code and rewrote service account token generation

users/users/keycloak/schemas.py
khyurri: This code was refactored

README.md
```diff
@@ -19,31 +21,65 @@ For now, BadgerDoc is working with vectorized and scanned documents, but it has
...
-docker build infra/docker/python_base -t 818863528939.dkr.ecr.eu-central-1.amazonaws.com/badgerdoc/python_base:0.1.7
+docker-compose -f docker-compose-dev.yaml up -d
 ```
-P.S. You can get this tag from any microservice Dockerfile like ```processing/Dockerfile```
 
-#
+## Keycloak local configuration
+_It's a good idea to automate this section_
+
+Important! This not secure configuration, follow KeyCloak best practices to
```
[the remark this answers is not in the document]
khyurri: Updated

README.md
khyurri: Updated

lib/tenants/src/dependency.py
```diff
@@ -8,12 +10,16 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(os.getenv("LOG_LEVEL", "INFO"))
+
 
 class TenantDependencyBase:
     def __init__(
         self,
         key: str = "",
         algorithm: str = "RS256",
+        # TODO: Fix type from bagerdoc to badgerdoc
```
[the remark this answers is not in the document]
khyurri: Fix typo

users/users/keycloak/utils.py
```diff
@@ -18,19 +15,6 @@ def create_filters(users: Users) -> Dict[str, Union[str, None]]:
...
-            if e.status in {401, 403}:
-                Logger.info(
-                    "Access Token for ClientViewsAllUsers_Role is expired. "
-                    "Creating new one"
-                )
-                await access_token.request_new_token()
-                await asyncio.sleep(1)
-            else:
-                raise e
+async def get_username(user_id: str):
```
cakeinsauce, line 18: May you specify return typing?

cakeinsauce (approved):

PR #451, opened 2023-03-21
feat: add uui compatible pagination to filter lib

spiridonovfed:


lib/filter_lib/tests/test_pagination.py
```diff
@@ -7,7 +7,9 @@
     make_pagination,
     paginate,
 )
+from ..src.query_modificator import form_query
```
khyurri, line 10: Avoid using relative imports, if it's possible

khyurri (changes requested):
Please, add filtering logic

lib/filter_lib/src/schema_generator.py
```diff
@@ -60,9 +60,26 @@ class _Sorts(GenericModel, Generic[TypeT]):
...
 
+    @root_validator
+    def root_validate(  # pylint: disable=no-self-argument
+        cls, values: Any
+    ) -> Any:
+
+        page_num = values.get("page_num")
+        page_offset = values.get("page_offset")
+
+        if page_num:
```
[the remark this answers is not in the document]
khyurri: Or:
```
if all((page_num, page_offset,)): 
    # both on
elif page_num or page_offset:
    # one is on
# both off
```

lib/filter_lib/tests/test_pagination.py
spiridonovfed: All imports in library itself and in tests for it are written in a relative manner.
I would prefer to keep all imports this way in order not to break something unintentionally

lib/filter_lib/src/schema_generator.py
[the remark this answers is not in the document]
spiridonovfed: Good point. Rewritten like the first option

lib/filter_lib/tests/test_query_modifier.py
```diff
@@ -1,3 +1,7 @@
+from random import randint
```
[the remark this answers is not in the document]
spiridonovfed: Corrected

lib/filter_lib/src/pagination.py
```diff
@@ -9,24 +9,48 @@
...
 ) -> Tuple[Query, PaginationParams]:
-    has_more: bool = False
-    max_count = page_size * 10 + 1
-    total_results: int = query.limit(max_count).count()
-    if total_results == max_count:
-        has_more = True
-        total_results -= 1
-    query = query.limit(page_size)
-    query = query.offset((page_number - 1) * page_size)
+    if page_number is not None:
```
khyurri, line 26: Please, simplify this code, you have got 2 branches with similar algorithm

lib/filter_lib/src/pagination.py
```diff
@@ -35,11 +59,18 @@ def _calculate_num_pages(page_size: int, total_results: int) -> int:
...
 
 def paginate(items: Sequence[T], pag_params: PaginationParams) -> Page[T]:
-    pag = PaginationOut(
-        page_num=pag_params.page_num,
-        page_size=pag_params.page_size,
-        min_pages_left=pag_params.min_pages_left,
-        total=pag_params.total,
-        has_more=pag_params.has_more,
-    )
+    pagination_out_attrs = {
```
khyurri, line 62: Use dict comprehension to build params for PaginationOut

lib/filter_lib/src/query_modificator.py
```diff
@@ -70,7 +70,10 @@ def form_query(
         pagination = Pagination(page_num=1, page_size=15).dict()
     try:
         query, pag = make_pagination(
-            query, pagination["page_num"], pagination["page_size"]
+            query,
+            pagination.get("page_num"),
```
khyurri, line 74: Unify access to keys: use `[...]` or `.get("...")`

lib/filter_lib/src/schema_generator.py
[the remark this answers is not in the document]
khyurri: Second one is more straightforward

lib/filter_lib/src/pagination.py
spiridonovfed: The logic in those two branches differ.

When using 'page_number' (old way):
offset = (page_number - 1) * page_size
'total_results' is calculated for 10 pages only.
'has_more' shows whether there are more entries that are in first 10 pages.

When using 'page_offset' (new_functionality):
offset = page_offset passed in (page_num is not used in calculation as above)
'total_results' is calculated for all the results that satisfy query filters - if the logic should be different, please write down what should 'total_results' mean in this particular case
'has_more' shows whether there are more entries after the "limit" == "page_size". No logic about 10 pages

lib/filter_lib/src/pagination.py
spiridonovfed: corrected with using dict comprehension

lib/filter_lib/src/query_modificator.py
spiridonovfed: unified to using .get()

lib/filter_lib/tests/test_pagination.py
spiridonovfed: All relative imports are rewritten to absolute

lib/filter_lib/src/pagination.py
```diff
@@ -35,11 +63,9 @@ def _calculate_num_pages(page_size: int, total_results: int) -> int:
...
+        field_name: getattr(pag_params, field_name)
+        for field_name in pag_params._fields
```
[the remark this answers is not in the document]
spiridonovfed: I do that because I can iterate through named_tuple attributes to create dict with dict comprehension and unpack it to PaginationOut

lib/filter_lib/src/schema_generator.py
```diff
@@ -60,9 +60,29 @@ class _Sorts(GenericModel, Generic[TypeT]):
 
 
 class Pagination(BaseModel):
-    page_num: int = Field(1, gt=0)
+    page_num: Optional[int] = Field(None, gt=0, le=100)
```
khyurri, line 63: Why do you limit maximum `page_num`?

spiridonovfed: fixed - this constraint shoud be applied to page_offset

Issue #454, opened 2023-03-21
Incorrect work of the button "Edit" job


Iogsotot:
**Step to reproduce:** 
1) go to **Extraction** tab
2) go to any **Job** page
3) click on **"Edit"** job button 

**Actual result:**
Page with navigation component, but without Edit job form
![image](https://user-images.githubusercontent.com/50149163/226627826-c10b515c-4d90-4961-baf8-6d97ebc61d8d.png)

**Expected result:** 
 Edit job form and job edit functionality
 
 *Network:
![image](https://user-images.githubusercontent.com/50149163/226628325-5fa8e356-298b-4026-925b-342db75abb34.png)
![image](https://user-images.githubusercontent.com/50149163/226628399-014b90c6-edb5-4986-b8e1-6469f57191a9.png)
![image](https://user-images.githubusercontent.com/50149163/226628488-1ec011c2-fc79-43eb-8b1c-c7ab90d41e55.png)

Iogsotot:
Possible duplicate of https://github.com/epam/badgerdoc/issues/443

serereg:
```
INFO:     10.233.96.227:49382 - "POST /api/v1/pipelines/pipelines/search HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 375, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/usr/local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/fastapi/applications.py", line 208, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc
  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/usr/local/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
[...]

PR #464, opened 2023-03-23
Fix: fix SQLAlchemy version

serereg:
Closes https://github.com/epam/badgerdoc/issues/454


pipelines/pyproject.toml
```diff
@@ -10,7 +10,7 @@ python-dotenv = "^0.19.0"
 fastapi = "^0.70.0"
 pydantic = "^1.8.2"
 psycopg2-binary = "^2.9.1"
-sqlalchemy = "^1.3.24"
+sqlalchemy = "~1.3.24"
```
khyurri, line 13: Let's bind to exact version `SQLAlchemy = "1.3.24"` 

Also, may I ask you to bind in all other services, lock and test?

serereg: Now version of SQLAlchemy in all services is ==1.3.24, except scheduler, because of its tests: they fail with 1.3.24.

Issue #370, opened 2023-02-24
Label tags overlap when there is not enough space between them


MarinaVorobeva21:
**Setup**
An annotation job is created and started.
**Steps**

- Create a few intersecting labels over a chunk of text such that ends of corresponding labels spans are close to one another (several characters).
- Label tags (little rectangles above annotated text) obscure one another

![image](https://user-images.githubusercontent.com/98399603/221183035-dd11af04-0def-4d6e-99c4-5a32710bce1f.png)

MikhailFokanov:
@khyurri, AFAIK the same strategy, that is implemented for exactly the same offsets should be done. The "if" condition, that labels override each other should be enhanced so that if the last token of the selection is "close" (there should be some predefined distance) to the other than the labels should shown as it has been done for fully overlapping annotations.

PR #469, opened 2023-03-24
#370 Added possibility to hide/show sidebars & fixed some styles

nanbratan:


web/src/components/task/task-sidebar-flow/index.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { FC, useEffect, useMemo, useState } from 'react';
+import { Button, FlexRow, Panel, TabButton } from '@epam/loveship';
+import styles from './styles.module.scss';
```
Iogsotot, line 3: It would be better to move the styles to the bottom of the list of imports to avoid possible overwriting problems from other components. Unfortunately, now not all styles are isolated inside their component and it is important to follow the import order

Iogsotot (approved):

web/src/components/task/task-sidebar-flow/index.tsx
nanbratan: I'm not sure that I understand how any global styles may override our incapsulated(module) styles

PR #480, opened 2023-03-27
fix: change job editing behavior

### LINKED ISSUE #425: Change job editing behavior
At the moment, after editing the job, the results of the tasks by users who were removed from this job are not saved. Their tasks are completely deleted.

We need to keep finished users annotations and validations.

Andrka:
Closes #425

### DIFF
--- a/annotation/annotation/jobs/services.py
+++ b/annotation/annotation/jobs/services.py
@@ -1,4 +1,4 @@
-from collections import defaultdict
+from collections import Counter, defaultdict
 from typing import Any, DefaultDict, Dict, List, Optional, Set, Tuple, Union
 from uuid import UUID
 
@@ -193,7 +193,7 @@ def update_jobs_users(
     validation_type: ValidationSchema,
     tenant: str,
     is_manual: bool = True,
-) -> Optional[Set[UUID]]:
+) -> Tuple[Set[UUID], Set[UUID]]:
     new_annotators = patch_data.get("annotators", [])
     new_validators = patch_data.get("validators", [])
     new_owners = patch_data.get("owners", [])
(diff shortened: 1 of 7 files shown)


annotation/annotation/jobs/services.py
```diff
@@ -555,3 +581,139 @@ def update_jobs_names(db: Session, jobs_names: Dict):
...
+
+    return set(annotation_tasks_to_delete + validation_tasks_to_delete)
+
+
+Task = Dict[str, Union[int, List[int], bool]]
+
+
+def remove_pages_in_work(tasks: List[Task], pages_in_work: FileInWork):
+    """Remove already started pages from tasks. Removing inplace."""
+    tasks.sort(key=lambda x: x["pages"])
```
spiridonovfed, line 662: Looks like the function does also an inplace sorting, which is not mentioned in function name and docstring

annotation/annotation/jobs/services.py
```diff
@@ -228,7 +230,28 @@ def update_jobs_users(
...
-    return deleted_users
+    return deleted_users, saved_users
+
+
+def get_users_to_save(
+    job: Job, new_job_data: dict, saved_users: set
+) -> Tuple[Set[UUID], Set[UUID]]:
+    old_annotators, old_validators = set(
+        annotator.user_id for annotator in job.annotators
+    ), set(validator.user_id for validator in job.validators)
```
spiridonovfed, line 241: You can use set comprehension to achieve the same result

spiridonovfed (changes requested):

spiridonovfed (approved):

Issue #488, opened 2023-03-29
Add reason why validation can't be confirmed


zefiris87:
1. Create hierarchical validation with 2 or more annotators
2. Finish annotations
3. Start validation, validate just one page
4. Try to finish validation (following popup appears)

exp.: error message appears, user get instructions to validate all pages

act.: nothing is happening when user clicks on the "confirm" button

![image](https://user-images.githubusercontent.com/124054458/228539471-1ad501d2-51b1-42af-b179-92ad5dca4b71.png)

Iogsotot:
Fixed the logic for validation: now the button finish validation is only available if all pages have been marked as valid or invalid, so the wronng modal won't appear

MarinaVorobeva21:
Fixed

PR #493, opened 2023-03-30
feat : finish validation#488

Iogsotot:
Closes https://github.com/epam/badgerdoc/issues/488

- [x] the button `finish validation` is only available if all pages have been marked as valid or invalid
- [x] added safe navigator operator for dataAttributes reading


web/src/shared/hooks/use-annotations-taxons.ts
```diff
@@ -15,11 +15,11 @@ export default function useAnnotationsTaxons(annotationsByPages?: PageInfo[]): M
             for (let page of annotationsByPages) {
                 for (let obj of page.objs) {
                     if (
-                        obj.data.dataAttributes?.[0] &&
-                        obj.data.dataAttributes?.[0].value &&
-                        obj.data.dataAttributes?.[0].type === 'taxonomy'
+                        obj.data?.dataAttributes?.[0] &&
```
nanbratan, line 18: I think it may be better to not check whole path every time and do somethig like this:

```
for (let obj of page.objs) {
    сonst firstAttribute = obj.data?.dataAttributes?.[0];
    
    if (
        firstAttribute?.value &&
        firstAttribute?.type === 'taxonomy'
    ) {
        taxonIdArr.push(firstAttribute.value);
    }
}
```

Or even this:
```

for (let obj of page.objs) {
    сonst { value, type } = obj.data?.dataAttributes?.[0] || {};
    
    if (value && type === 'taxonomy') {
        taxonIdArr.push(value);
    }
}
```

nanbratan (approved):

PR #499, opened 2023-03-31
feat(actions): add pull request title linter

### LINKED ISSUE #7: Add Conventional Commits check to GitHub Actions
As a developer I want to check that contributors follows convention of commits naming in the repository. We should add [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) check to the CI. Here is a good linter for this purpose written in python which supports pre-commit integration as well: [commitizen](https://github.com/commitizen-tools/commitizen)

DOD:
- [ ] Added commit messages check step in CI

cakeinsauce:
Closes #7

### DIFF
--- a/.github/workflows/pr-lint.yml
+++ b/.github/workflows/pr-lint.yml
@@ -1,0 +1,16 @@
+name: lint pull request
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  semantic_pr:
+    name: Validate PR title
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v5
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


khyurri:
@cakeinsauce 
Thank you for PR, please, check this discussion https://github.com/epam/badgerdoc/issues/7 
Should we use https://github.com/amannn/action-semantic-pull-request instead of your solution?

cakeinsauce:
> @cakeinsauce Thank you for PR, please, check this discussion #7 Should we use https://github.com/amannn/action-semantic-pull-request instead of your solution?
@khyurri We could use it as well, I'm just not comfortable using third parties GH actions.
Mine and https://github.com/amannn/action-semantic-pull-request do the same stuff. We use [commitlint](https://github.com/conventional-changelog/commitlint)
And I'm not sure, but seems like https://github.com/amannn/action-semantic-pull-request only checks PR title according to conventional commits spec, and not commits themself

khyurri (approved):

PR #515, opened 2023-04-10
refactor: Removed not used return values from context. Simplify logic.

### LINKED ISSUE #387: Front-end refactoring
- [ ] Task annotator context

nanbratan:
Closes #387


web/src/api/typings/tasks.ts
```diff
@@ -57,4 +55,4 @@ export type TaskStatus = 'Pending' | 'Ready' | 'In Progress' | 'Finished';
 
 export type ValidationPageStatus = 'Valid Page' | 'Invalid Page';
 
-export type TTaskUsers = Record<'owners' | 'validators' | 'annotators', TUserShort[]>;
+// export type TTaskUsers = Record<'owners' | 'validators' | 'annotators', TUserShort[]>;
```
Iogsotot, line 58: pls, delete commented code

web/src/components/task/task-modal/task-validation-modal.tsx
```diff
@@ -45,16 +44,17 @@ export const FinishTaskValidationModal: FC<IProps> = (modalProps) => {
...
         abort,
         validSave,
         success,
         onRedirectAfterFinish,
         onSaveForm
     } = modalProps;
+
     const [validatorUserId, onValidatorUserIdChange] = useState(
-        allUsers.validators[0]?.id || allUsers.annotators[0]?.id
+        allUsers.validators[0] || allUsers.annotators[0]?.id
```
Iogsotot, line 57: Please help me understand why you are leaving out the id in this part of the code? changed the field object `allUsers.validators[0]`?

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -145,12 +130,12 @@ type ContextValue = SplitValidationValue &
...
         setSelectedAnnotation: (annotation: Annotation | undefined) => void;
         selectedLabels?: Label[];
-        onLabelsSelected: (labels: Label[], pickedLabels: string[]) => void;
+        onLabelsSelected: (labels: Label[]) => void;
         setSelectedLabels: (labels: Label[]) => void;
         latestLabelsId: string[];
         isDocLabelsModified: boolean;
         getJobId: () => number | undefined;
-        linksFromApi?: DocumentLink[];
+        // linksFromApi?: DocumentLink[];
```
Iogsotot, line 138: pls delete commented code

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -252,39 +232,12 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
...
-    const getJobId = (): number | undefined => (task ? task.job.id : jobId);
-
-    const getFileId = (): number | undefined => (task ? task.file.id : fileMetaInfo?.id);
-    const { isOwner, taskUsers } = useTaskUsers(task);
-
+    const { data: task, isLoading: isTaskLoading, refetch: refetchTask } = useTaskById({ taskId });
     const { data: job } = useJobById({ jobId: task?.job.id });
 
-    let pageNumbers: number[] = [];
+    const getJobId = (): number | undefined => (task ? task.job.id : jobId); // Do we need this???
```
Iogsotot, line 238: @sbilevich 
could you share your opinion?

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -407,41 +376,32 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
...
+                [pageNum]: allAnnotations[pageNum].filter((ann) => {
                     if (
-                        anntn &&
-                        anntn.children &&
-                        anntn.boundType === 'table' &&
-                        (anntn.children as number[]).includes(+ann.id) &&
+                        annotation &&
+                        annotation.children &&
+                        annotation.boundType === 'table' &&
+                        (annotation.children as number[]).includes(+ann.id) &&
```
Iogsotot, line 392: Many thanks!

web/src/connectors/task-annotator-connector/use-validation.tsx
```diff
@@ -59,18 +50,17 @@ export type ValidationValues = {
     onCancelClick: () => void;
     onClearTouchedPages: () => void;
     onAddTouchedPage: () => void;
-    setValidPages: (pages: number[]) => void;
+    setValidPages: (pages: number[]) => void; // -
```
Iogsotot, line 53: ?

web/src/connectors/task-annotator-connector/use-validation.tsx
```diff
@@ -59,18 +50,17 @@ export type ValidationValues = {
...
     onFinishValidation: () => void;
     onSaveEditClick: () => void;
-    setAnnotationSaved: Dispatch<SetStateAction<boolean>>;
+    setAnnotationSaved: Dispatch<SetStateAction<boolean>>; // -
```
Iogsotot, line 56: ?

Iogsotot:
The PR is very large and touches on many key points in the application (Context), I think it will be safer to get a second review from Sveta @sbilevich

web/src/components/task/task-modal/task-validation-modal.tsx
nanbratan: now we have validators from job, field "job.validators" contains string[]

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -407,41 +375,32 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
     };
 
     const deleteAnnotation = (pageNum: number, annotationId: string | number) => {
-        const pageAnnotations = allAnnotations[pageNum] ?? [];
-        const anntn: Maybe<Annotation> = pageAnnotations.find((el) => el.id === annotationId);
-        if (anntn?.labels) {
```
sbilevich, line 412: is this logic moved somewhere or just deleted?

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -252,39 +231,12 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
...
-        isTaskLoading = result.isLoading;
-        refetchTask = result.refetch;
-    }
-
-    const getJobId = (): number | undefined => (task ? task.job.id : jobId);
-
-    const getFileId = (): number | undefined => (task ? task.file.id : fileMetaInfo?.id);
-    const { isOwner, taskUsers } = useTaskUsers(task);
-
+    const { data: task, isLoading: isTaskLoading, refetch: refetchTask } = useTaskById({ taskId });
```
sbilevich, line 234: maybe should be enabled only if taskId?

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -210,12 +192,11 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
     const [selectionType, setSelectionType] = useState<
         AnnotationBoundType | AnnotationLinksBoundType | AnnotationImageToolType
     >('free-box');
-    const [selectedTool, setSelectedTool] = useState<AnnotationImageToolType>('pen');
+    const [selectedTool, setSelectedTool] = useState<AnnotationImageToolType>(ToolNames.pen);
     const [selectedAnnotation, setSelectedAnnotation] = useState<Annotation | undefined>();
-    const [isDataTabDisabled, setIsDataTabDisabled] = useState<boolean>(dataTabDefaultDisableState);
```
sbilevich, line 215: how we disable Data tab now?

nanbratan: We don't do it anymore as `isDataTabDisabled` is not using anywhere

nanbratan: `isDataTabDisabled` wasn't using anywhere, so we don't need to disable it at all

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
nanbratan: it's already enabled only if taskId is exists
```
export const useTaskById: QueryHookType<{ taskId?: number }, Task> = ({ taskId }) =>
    useQuery(
        ['task', taskId],
        async () =>
            useBadgerFetch<ApiTask>({
                url: `${namespace}/tasks/${taskId}`,
                method: 'get'
            })(),
        { select: mapTaskFromApi, enabled: Boolean(taskId) }
    );
    ```

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
sbilevich: The purpose of labels attr is to handle a case of annotation intersection. If 2 annotations are one the same place(or just ended in the same place) we show labels one above another, so when we delete one of them we still need to show another

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
sbilevich: sorry, missed it

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
sbilevich: But we still need to disable Data tab if there is no selected annotation, right?

nanbratan: As I said - we don't use `isDataTabDisabled`, we didn't use it before this refactoring and we don't use it now.

Please refer to place where we use `isDataTabDisabled`, I couldn't find any place with it

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
nanbratan: What do you think old code did? Could you explain it please?

As I understand:

We get the annotation by annotationId and pageNum from allAnnotations
Then we remove label from the annotation's labels.
Then we remove the annotation by annotationId and pageNum from allAnnotations
What was the point of removing label from annotation.labels if we're going to remove this annotation at all?
We just mutated object and them removed it from array.
There is no point to mutate object if we are going to delete it.

sbilevich: image
in this case when you deleting label, you delete annotation but also you should remove deleted label from labels attr of other annotations.

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
sbilevich: Ok, I see. In that case could you please check how it disables now. If it works fine - ok, we can delete this, but if it doesn't work as expected, maybe we still need this code?

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
nanbratan: Was a code, which I removed, deleting labels of other annotations?
As I understand it was changing only its own labels array. I see how we get annotation by annotationId and modify it, I don't see how we get all other annotations and modify its labels.

if we need behaviour which you're describing - I suggest you to create bug/improvement to implement it as it looks like that it wasn't implemented.

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
nanbratan: It looks like we don't disable "Data" tab at all
image

Also I don't think that we need to keep not used code that may be useful at some point in the future, we can return it if we need to

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
sbilevich: Labels are grouped by ending token and stored separately from annotations in map `tokenLabelsMap`.
So when you add or delete annotation you need to manage those labels separately, and deleting of annotation isn't enough.
I agree, that it not the clearest solution, but you need to take it into account.
I tried a case with many annotations ending on the same token on main and it works fine, but it doesn't work in your branch.

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
nanbratan: Because code which may be useful at some point in future === useless code which definitely won't be useful in future

Because of such approach we will have garbage in code-base
And about `If you remove this now and other developer will need to fix it` - it's not a problem to implement it again, but having garbage in code base is a problem

sbilevich (approved):

PR #521, opened 2023-04-13
Fix/labelstudio export

### LINKED ISSUE #66: Add converter from bbox to offsets in export and import
Add addition converter to export/import between PDF and plain text format. 

BadgerDoc is currently working with PDF and bbox, we need also support plain text with offsets as source format (converter) and export PDF + bbox to plain text with offsets

Part 2: 
- [x] From Label Studio format converter must create bdocs manifest and fill database (AnnotationDocs?)

serereg:
Closes #66


convert/convert/config.py
```diff
@@ -4,14 +4,15 @@
...
 from botocore.client import BaseClient
-from convert import logger
 from dotenv import load_dotenv
 from mypy_extensions import KwArg, VarArg
 from pydantic import BaseSettings, Field
 from requests import Session
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 
+from convert import logger
```
spiridonovfed, line 14: Here you import logger from another place.
But in the code down below you create several loggers in different modules inplace.

I think it should one way or another - create one logger instance and import it everywhere we need it or create different instances in every module if that's needed.
The first approach seems more clean

convert/convert/converters/labelstudio/badgerdoc_downloader.py
```diff
@@ -0,0 +1,132 @@
...
+        self,
+        s3_client: BaseClient,
+        s3_input_tokens: S3Path,
+        s3_input_manifest: S3Path,
+    ):
+        self.s3_input_tokens = s3_input_tokens
+        self.s3_input_manifest = s3_input_manifest
+        self.s3_client = s3_client
+
+    def download(
```
spiridonovfed, line 36: It's not only downloading relevant files, it also parses it.
I would call it "download_and_parse" or any other descriptive name

convert/convert/converters/labelstudio/badgerdoc_downloader.py
```diff
@@ -0,0 +1,132 @@
...
+        self, manifest: Manifest, tmp_dir
+    ) -> Dict[int, BadgerdocAnnotation]:
+        annotation_files = self.download_annotations(
+            manifest_s3_path=self.s3_input_manifest,
+            manifest=manifest,
+            tmp_dir=tmp_dir,
+        )
+        annotations = {}
+        for page_num, annotation_file in annotation_files.items():
+            annotations[int(page_num)] = AnnotationConverterToTheory(
```
spiridonovfed, line 73: what's the point of "int(page_num)". Do they come as strings?

convert/convert/converters/labelstudio/labelstudio_format.py
```diff
@@ -0,0 +1,350 @@
...
+                Annotation(
+                    id=self.DEFAULT_ID_FOR_ONE_ANNOTATION,
+                    result=objs + relations,
+                )
+            )
+
+        meta = Meta()
+        if badgerdoc_manifest:
+            meta = self.form_meta(badgerdoc_manifest, request_headers)
+
```
spiridonovfed, line 69: It looks weird a bit for me.
if badgerdoc_manifest, then we create "meta" and rewrite it in the next step.

maybe this way:
self.form_meta(badgerdoc_manifest, request_headers) if badgerdoc_manifest else Meta()

or:
if badgerdoc_manifest:
.....
else:
......

spiridonovfed (changes requested):

convert/convert/converters/labelstudio/badgerdoc_downloader.py
serereg: yes, they are stored is "1", "2", etc.

spiridonovfed (approved):

khyurri (approved):

Issue #121, opened 2022-12-22
Annotation fixtures amend testing data


cakeinsauce:
Fixture `annotation.tests.conftest.prepare_db_for_annotation_change_task_statuses` amend some data so next tests would fail. To reproduce place `annotation.tests.test_post_annotation.test_post_annotation_by_user_similar_doc_no_category` (just a random test that uses data from this fixture) to the bottom of its module.
You would get a similar error:
```shell
tests/conftest.py:244:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/conftest.py:188: in add_objects
    db.merge(obj)
venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py:2166: in merge
    return self._merge(
venv/lib/python3.8/site-packages/sqlalchemy/orm/session.py:2306: in _merge
    prop.merge(
venv/lib/python3.8/site-packages/sqlalchemy/orm/properties.py:287: in merge
    impl.set(dest_state, dest_dict, value, None)
venv/lib/python3.8/site-packages/sqlalchemy/orm/attributes.py:867: in set
    old = self.get(state, dict_, PASSIVE_RETURN_NEVER_SET)
venv/lib/python3.8/site-packages/sqlalchemy/orm/attributes.py:725: in get
    value = state._load_expired(state, passive)
venv/lib/python3.8/site-packages/sqlalchemy/orm/state.py:652: in _load_expired
    self.manager.deferred_scalar_loader(self, toload)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

mapper = <Mapper at 0x104281700; ManualAnnotationTask>
state = <sqlalchemy.orm.state.InstanceState object at 0x104acc970>
attribute_names = {'deadline', 'file_id', 'id', 'is_validation', 'job_id', 'pages', ...}

    def load_scalar_attributes(mapper, state, attribute_names):
        """initiate a column-based attribute refresh operation."""

        # assert mapper is _state_mapper(state)
        session = state.session
        if not session:
[...]

PR #573, opened 2023-04-25
fix: work of next button#554

### LINKED ISSUE #554: [MAJOR] How the Next document button works
If the annotator has tasks from different jobs and there is a change of priorities (ex. from UAT 12.6 job with 643 files to UAT 12.6 primary job with 60 files), then he can't start from the "last document" of the new job, because then there is no next document for the Next button and the annotator needs to go to Dashboard for a new task
image
image
The workaround only:
Since there is no filtering by Job, it needs to manually scroll through the pages to find the start of the other job.
image

serereg:
Closes #554

### DIFF
--- a/annotation/tests/test_tasks_crud_cr.py
+++ b/annotation/tests/test_tasks_crud_cr.py
@@ -40,6 +40,7 @@ ANNOTATORS = (
     ),
     User(user_id="5c9a1d6f-63ab-4777-bc1c-53984a5c9105"),
     User(user_id="9eace50e-613e-4352-b287-85fd91c88b51"),
+    User(user_id="1cace611-713f-5353-c268-95ff92c87b62"),
 )
 SEARCH_TASKS_PATH = CRUD_TASKS_PATH + "/search"
 CRUD_CR_JOBS = (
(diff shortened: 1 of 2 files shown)


spiridonovfed (approved):

spiridonovfed:
Code changes are approved, but please correct tests

annotation/tests/test_tasks_crud_cr.py
```diff
@@ -1449,20 +1505,22 @@ def test_search_tasks_pagination(
...
         (
             "user_id",
             "eq",
             "9eace50e-613e-4352-b287-85fd91c88b51",
             list(range(3, 7)),
         ),
-        ("is_validation", "ne", False, [5, 6, 20, 23]),
-        ("is_validation", "ne", False, [5, 6, 20, 23]),
+        # FIXME
+        pytest.param("is_validation", "ne", False, [5, 6, 20, 23], marks=pytest.mark.xfail(reason="fix the test")),
```
serereg, line 1517: It relates to 
https://github.com/epam/badgerdoc/issues/121

docker-compose-dev.yaml
```diff
@@ -95,6 +95,9 @@ services:
     depends_on:
       badgerdoc-postgresql:
         condition: service_healthy
+    volumes:
+      - ./annotation/:/opt/annotation/
+    command: bash -c "alembic upgrade head && uvicorn annotation.main:app --host 0.0.0.0 --port 8080 --reload"
```
cakeinsauce, line 100: We have this line in Dockerfile https://github.com/epam/badgerdoc/blob/main/annotation/Dockerfile#L20
Just specify build target
```yml
    build:
      context: .
      target: build-dev
```

cakeinsauce: And do you think we need to mount the volume? We copy all necessary files during build in the Dockerfile

serereg: Done

annotation/tests/test_tasks_crud_cr.py
serereg: Done

PR #583, opened 2023-04-27
fix: fix 'total' calculation in filter_lib and relevant tests

spiridonovfed:


serereg (approved):

lib/filter_lib/src/pagination.py
```diff
@@ -28,18 +28,23 @@ def make_pagination(
...
-    has_more = total_results == max_count
 
-    if total_results == max_count:
-        total_results -= 1
+    total_results_query = initial_query.limit(max_count)
+    if not page_num:  # "page_offset" is used instead of "page_num"
+        total_results_query = total_results_query.offset(page_offset)
+    total_results_count = total_results_query.count()
+
+    has_more = total_results_count == max_count
```
serereg, line 37: Is the situation where total_results_count == max_count == 0 proceeded properly?

lib/filter_lib/src/pagination.py
```diff
@@ -28,18 +28,23 @@ def make_pagination(
...
+    if total_results_count == max_count:
```
serereg, line 38: Is this case total_results_count == max_count == 0 proceeded properly?

spiridonovfed: It's not a valid case, because 'max_count' is always > 0 for sure. It is calculated one level above

PR #585, opened 2023-04-28
feat: add tasks distribution option

### LINKED ISSUE #538: [CRITICAL] Splitting a multipage document into several different Tasks
The original doc has 50 pages.
The document was splitted into 2 tasks not exceeding 45 pages per user, and instead of the same document assignment between 3 users, which
is a project necessity, the same document assignment was performed between 4 users with random parts of the document in tasks.
image
image
image
image

Andrka:
Closes #538


annotation/annotation/distribution/main.py
```diff
@@ -72,6 +77,8 @@
 )
 from annotation.tasks import create_tasks as create_db_tasks
 
+dotenv.load_dotenv(dotenv.find_dotenv())
+SPLIT_MULTIPAGE_DOC = os.getenv("SPLIT_MULTIPAGE_DOC", "false")
```
serereg, line 81: I would use bool variable instead of str
like 
`SPLIT_MULTIPAGE_DOC = True if os.getenv("SPLIT_MULTIPAGE_DOC", "false") == "true" else False`

serereg (approved):

PR #601, opened 2023-05-05
fix: fix validation schema for extensive coverage

### LINKED ISSUE #569: update BE api/v1/annotation/annotation/{task_id} for extensive coverage
When creating a job type: extensive coverage, task: validation, when finishing validation, it is necessary to take into account the possibility of transferring data to the backend for validation and new annotations at the same time (now BE return 422 error).

Actual logic:
if FE send request with this payload
image
BE response
image

It now works with a workaround with 2 different requests (with data for pages field and with data for validation field separately) and creates 2 revisions instead of 1 revision.

- [ ] update BE  `api/v1/annotation/annotation/{task_id}`
- [ ] update FE to use the new logic

serereg:
Closes #569

### DIFF
--- a/annotation/.env
+++ b/annotation/.env
@@ -1,10 +1,10 @@
 APP_HOST="0.0.0.0"
 APP_PORT=8080
-POSTGRES_HOST="badgerdoc-postgresql"
+POSTGRES_HOST=localhost
 POSTGRES_PORT=5432
 POSTGRES_USER="postgres"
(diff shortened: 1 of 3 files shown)


spiridonovfed (changes requested):
This is new functionality and I am sure that we should add 1-2 tests to cover it

spiridonovfed (approved):

Issue #555, opened 2023-04-18
[MAJOR] No filtering by Job Name


MarinaVorobeva21:
It's necessary for the absence of blocks to create several Jobs

MarinaVorobeva21:
1. There is no grouping of the names in the list of filter
![image](https://github.com/epam/badgerdoc/assets/98399603/78c35d88-72aa-4067-8923-60a950d3be71)
2. Filtering is not working on other pages from the first one. (Go to the 2nd page and try to filter)
https://github.com/epam/badgerdoc/assets/98399603/190cb253-54ab-48d1-97ca-4ce32c957b77
3. Filtering is not working for all the records (e.g. try to choose job names from the list)
https://github.com/epam/badgerdoc/assets/98399603/6867e728-479b-45a1-b453-43f0c9bf0b1d
4. Sometimes not all the names(records) are displayed in the list. Refreshing the page is needed to see all the records.
5. Mouse wheel scrolling and cursor movement scrolling conflicts

MarinaVorobeva21:
Filtering is not working on the different from the 1st page.
image

The records are not displayed for all chosen filters (e.g. choose 66_1 in Job name filter or 001.jpg in the file name filter)

sokolmask:
Smartfilter is not in scope, consider as ready

PR #645, opened 2023-05-25
feat: Implemented filtering by file name and job name

### LINKED ISSUE #557: [MAJOR] No filtering by File name
No filtering by document name in the dashboard if they need to return to a specific document.

Tokhirov-Abzal:
Closes #557 and #555

### DIFF
--- a/web/src/connectors/tasks-table-connector/use-job-filter.ts
+++ b/web/src/connectors/tasks-table-connector/use-job-filter.ts
@@ -1,6 +1,5 @@
 import { useLazyDataSource } from '@epam/uui';
-import { jobsFetcher } from 'api/hooks/jobs';
-import { Job } from 'api/typings/jobs';
+import { jobPropFetcher } from 'api/hooks/jobs';
 import { useRef } from 'react';
 import { useColumnPickerFilter } from 'shared/components/filters/column-picker';
 import { createPagingCachedLoader } from 'shared/helpers/create-paging-cached-loader';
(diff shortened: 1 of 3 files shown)


web/src/connectors/tasks-table-connector/use-job-filter.ts
```diff
@@ -18,29 +17,30 @@ export const useJobFilter = ({ fieldName }: Props) => {
...
     );
 
-    const renderJobFilter = useColumnPickerFilter<Job, string, unknown, string>(
+    const renderJobFilter = useColumnPickerFilter<string, string, unknown, string>(
         jobNames,
         fieldName,
         {
             showSearch: true,
-            getName: (item) => (typeof item === 'boolean' ? String(item) : item.name)
+            getName: (item) => (typeof item === 'boolean' ? String(item) : item)
```
sbilevich, line 43: I believe we don't need getName here, as we have only names. Please try without it

web/src/connectors/tasks-table-connector/use-name-filter.ts
```diff
@@ -18,30 +17,30 @@ export const useNameFilter = ({ fieldName }: Props) => {
...
     );
 
-    const renderNameFilter = useColumnPickerFilter<FileDocument, string, unknown, string>(
+    const renderNameFilter = useColumnPickerFilter<string, string, unknown, string>(
         documentNames,
         fieldName,
         {
             showSearch: true,
-            getName: (item) => (typeof item === 'boolean' ? String(item) : item.original_name)
+            getName: (item) => (typeof item === 'boolean' ? String(item) : item)
```
sbilevich, line 43: The same

sbilevich (approved):

web/src/connectors/tasks-table-connector/use-job-filter.ts
Tokhirov-Abzal: Thanks for pointing it out. Removed

web/src/connectors/tasks-table-connector/use-name-filter.ts
Tokhirov-Abzal: Removed

PR #685, opened 2023-06-08
Feat/pdfminer to fitz in convert#630

thinklab:


convert/convert/converters/pdf/pdf_converter.py
```diff
@@ -101,3 +102,53 @@ def _create_neighbor_bbox(self, current_bbox: Bbox) -> Bbox:
         new_x0, new_y0 = x1, y0
         new_x1, new_y1 = (x1 - x0) + x1, y1
         return new_x0, new_y0, new_x1, new_y1
+
+
+class PlainPDFToBadgerdocTokensConverterPytz:
```
serereg, line 107: As I understand it is needed to replase PlainPDFToBadgerdocTokensConverter class by this one

convert/convert/converters/pdf/pdf_converter.py
```diff
@@ -101,3 +102,53 @@ def _create_neighbor_bbox(self, current_bbox: Bbox) -> Bbox:
...
+    def __init__(self) -> None:
+        self.offset = 0
+        self.page_size: Optional[PageSize] = None
+
+    def _convert_span(self, span):  # type: ignore
```
serereg, line 112: It is necessary to remove all type ignores. In order to simplify the process of determining types, you can add a breakpoint here and in debug mode, find out the type of this variable.

PR #706, opened 2023-08-10
feat: add keycloak build for different platforms

cakeinsauce:


cakeinsauce:
@khyurri Would be nice to add this to build

Issue #708, opened 2023-08-16
Failed to make requests to "assets" service


iSokrat:
There is an issue with making requests to "assets" service.

Steps to reproduce:
Precondition:
Set WEB_CORS env var  to *  in `.env` files and apply it (most probably, `.env.example` should be updated as well).
Steps:
1. Make sure all containers are run.
2. Open locally in browser `localhost:8083`.
3. Open Dev Tools -> Network tab (all requests should be shown).
4. Login into the app.
5. Check `search` request.

Actual result:
1) Preflight `OPTIONS` request is failed (no cors setup for "assets" service)
2) In case if p.1 is fixed, "search" POST request is failed.

Expected result:
The issues p.1 and p.1 are fixed: "search" requests work fine (and the rest requests to this service as well).

About p.1 - need to add CORS-related handler to all requests for "assets" service like it's done for "users" service.
About p.2 After investigation with @khyurri, it appeared we have duplicated table names (`files`). To resolve this, we've just renamed it to `assets_files`. The goal of this issue to investigate & fix similar issues with table names.

PR #709, opened 2023-08-16
fix: rename "files" table to "assets_files"

iSokrat:
Additionally, CORS-related handler for "assets" service was added to support preflight requests.

The fix for the issue described here https://github.com/epam/badgerdoc/issues/708 (but that issue requires more investigation - there is a chance we may catch similar problem in other places).


khyurri (approved):

iSokrat:
Based on discussion with @khyurri - it's okay to merge with failed tests, they will be fixed later.

Issue #710, opened 2023-08-18
Update nodejs version


Ziprion:
Frontend part can not be built under nodejs 16+

PR #711, opened 2023-08-18
refactor: update nodejs version

Ziprion:
Closes [#710](https://github.com/epam/badgerdoc/issues/710)

### DIFF
--- a/web/src/index.tsx
+++ b/web/src/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import ReactDOM from 'react-dom';
+import { createRoot } from 'react-dom/client';
 import { Router } from 'react-router-dom';
 import { createBrowserHistory } from 'history';
 import { Snackbar, Modals } from '@epam/uui-components';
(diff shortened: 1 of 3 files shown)


web/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:14 as build
+FROM node:18 as build
```
iSokrat, line 1: Did you try to rebuilt this container and check that app works fine via 8083?

web/src/index.tsx
```diff
@@ -45,4 +45,4 @@ const RoutedApp = () => (
     </Router>
 );
 
-ReactDOM.render(<RoutedApp />, document.getElementById('root'));
+createRoot(document.getElementById('root')!).render(<RoutedApp />);
```
iSokrat, line 48: Am I right TS started to warn this place without this sign? (Because in old interface we didn't have this).

Ziprion: Yeap, it's from react documentation

Issue #712, opened 2023-08-18
Popup dialog does not display correctly


Ziprion:
Popup dialog does not display correctly

<img width="1414" alt="MicrosoftTeams-image" src="https://github.com/epam/badgerdoc/assets/74115008/8104e2aa-5f66-40ff-aac6-fb66b06ccac9">

PR #713, opened 2023-08-18
refactor: popup dialog does not display correctly

Ziprion:
Closes [#712](https://github.com/epam/badgerdoc/issues/712)

### DIFF
--- a/web/src/components/basement/basement-popup/basement-popup.module.scss
+++ b/web/src/components/basement/basement-popup/basement-popup.module.scss
@@ -10,13 +10,13 @@ $popupWidth: 45%;
   bottom: 0;
   background-color: $bg-color;
   opacity: 0.8;
-  z-index: 1;
+  z-index: 98;
 }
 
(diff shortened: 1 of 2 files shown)


web/src/components/basement/basement-popup/basement-popup.module.scss
```diff
@@ -10,13 +10,13 @@ $popupWidth: 45%;
...
   opacity: 0.8;
-  z-index: 1;
+  z-index: 98;
 }
 
 .popup {
   padding: 33px 36px 22px 25px;
   position: fixed;
-  z-index: 2;
+  z-index: 99;
```
iSokrat, line 19: Did you test that this value doesn't affect other placed except that one we fix?

Ziprion: Yeap, that component is used just in one place

iSokrat (approved):

PR #718, opened 2023-08-23
fix: values of convert host address & search cors

SunnyCapt:


assets/.env
```diff
@@ -16,5 +16,5 @@ GOTENBERG_FORMATS=[".txt",".docx",".doc",".bib",".xml",".csv",".fodt",".ltx",".o
 
 IMAGE_FORMATS=[".png",".bmp", ".pbm", ".pct", ".pgm", ".ppm", ".ras", ".tiff"]
 
-SERVICE_CONVERT_TXT="http://${CONVERT_SERVICE_HOST}/text/import"
-SERVICE_CONVERT_PDF="http://${CONVERT_SERVICE_HOST}/pdf/import"
+SERVICE_CONVERT_TXT="http://${CONVERT_SERVICE_HOST}:8080/text/import"
```
khyurri, line 19: May I suggest using {CONVERT_SERVICE_HOST} instead of http://${CONVERT_SERVICE_HOST}:8080 here?

This would allow for greater flexibility, as the port can be changed in different installations, and it would be more efficient to keep the service configuration in a single .env file.

Additionally, we could consider adding CONVERT_SERVICE_PORT and CONVERT_SERVICE_HTTP to the .env.example file and reusing them here.

Lastly, it might be beneficial to remove SERVICE_CONVERT_TXT and SERVICE_CONVERT_PDF from the assets/.env file and move them to the code, as this would enable us to use a single file configuration.

khyurri (approved):
Great!

PR #726, opened 2023-08-28
fix: refresh token request

Ziprion:


web/src/shared/helpers/auth-tools.ts
```diff
@@ -62,14 +63,26 @@ export const getCurrentTenant = (): string => localStorage.getItem(AUTH_CURRENT_
 
 export const refetchToken = async () => {
```
iSokrat, line 64: Am I right that there was some suggestion from Ruslan to increase time we refresh this token? Should this logic be updated somewhere or not?

iSokrat, line 64: I think after the fix we can remove that sentence we added in README about increasing of token lifespan (or we agreed to not do this, do you remember?).

Ziprion: [PR](https://github.com/epam/badgerdoc/pull/719) is not approved, so I will just close it

iSokrat: > It depends on keycloak's settings.

Yep, I just didn't remember what Ruslan proposed - either unbind from keycloak's setting or just increase this default timeout. My opinion is that would be nice to keep the current solution flexible depends on the settings in keycloak (but in this case we don't need to close your PR mentioned below).

iSokrat (approved):
LGTM, let's just share this idea that this timer can be adjusted in keycloak without touching this default timeout.

PR #733, opened 2023-09-08
perf: load annotations for PDF in chunks

iSokrat:
The goal of this PR is to improve performance of loading big PDF files (1000+) pages where we may potentially have up to 50 annotations. Unfortunately, this PR is quite big because during development there were a lot of different experiments which touched different parts of the app. It might be possible to split it into smaller PRs and merge them separately but this may require also a lot of time to re-test a lot of stuff for split functionality (it's easy to break something). Cos of this I propose to keep this solid PR.

In the future, it would be a good step to have refactoring mostly for the existing big react context because it contains tons of logic for different document modes  like "Validation only" / "Cross validation" etc. This modes may have own context/own hooks/own components.

### DIFF
--- a/web/src/connectors/task-annotator-connector/task-annotator-context.tsx
+++ b/web/src/connectors/task-annotator-connector/task-annotator-context.tsx
@@ -12,21 +12,19 @@ import { cloneDeep, isEqual } from 'lodash';
 import { Task, TTaskUsers } from 'api/typings/tasks';
 import { ApiError } from 'api/api-error';
 import {
+    AnnotationsResponse,
     DocumentLink,
-    useAddAnnotationsMutation,
-    useLatestAnnotations
+    useAddAnnotationsMutation
 } from 'api/hooks/annotations';
 import { useSetTaskFinishedMutation, useSetTaskState, useTaskById } from 'api/hooks/tasks';
 import { useCategoriesByJob } from 'api/hooks/categories';
 import { useDocuments } from 'api/hooks/documents';
 import { useJobById } from 'api/hooks/jobs';
-import { useTokens } from 'api/hooks/tokens';
 import {
     Category,
     CategoryDataAttributeWithValue,
     ExternalViewerState,
     FileDocument,
-    FilterWithDocumentExtraOption,
     Label,
     Link,
     Operators,
@@ -37,7 +35,12 @@ import {
 import { Job } from 'api/typings/jobs';
 import { FileMetaInfo } from 'pages/document/document-page-sidebar-content/document-page-sidebar-content';
 
-import { QueryObserverResult, RefetchOptions, RefetchQueryFilters } from 'react-query';
+import {
+    QueryObserverResult,
+    RefetchOptions,
+    RefetchQueryFilters,
+    UseQueryResult
+} from 'react-query';
 import {
     Annotation,
     AnnotationBoundType,
@@ -57,6 +60,7 @@ import useSyncScroll, { SyncScrollValue } from 'shared/hooks/use-sync-scroll';
 import { PageSize } from '../../shared/components/document-pages/document-pages';
 import {
     defaultExternalViewer,
+    getAnnotationsWithAppliedChanges,
     getCategoryDataAttrs,
     isValidCategoryType,
     mapAnnDataAttrs,
@@ -73,6 +77,10 @@ import { useNotifications } from 'shared/components/notifications';
 import { Text, Panel } from '@epam/loveship';
 import { getError } from 'shared/helpers/get-error';
 import { removeDuplicatesById } from './utils';
+import { useDocumentDataLazyLoading } from './use-document-data-lazy-loading';
+import { useDocumentDataFullLoading } from './use-document-data-full-loading';
+
+type WithFunctionStub<T> = T | (() => void);
 
 type ContextValue = SplitValidationValue &
     SyncScrollValue &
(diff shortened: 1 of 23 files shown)


web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -223,6 +254,9 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
     >({});
     const [externalViewer, setExternalViewer] =
         useState<ExternalViewerState>(defaultExternalViewer);
+    const [selectedRelatedDoc, setSelectedRelatedDoc] = useState<FileDocument | undefined>(
```
iSokrat, line 257: This was extracted from "useDocumentLinks" custom hook forcedly until we split the context into different parts.

web/src/connectors/task-annotator-connector/use-document-data-lazy-loading/index.ts
```diff
@@ -0,0 +1,114 @@
+import { useCallback, useEffect, useState } from 'react';
+import { useAnnotationsLazyLoading } from './use-annotations-lazy-loading';
+import { useTokensLazyLoading } from './use-tokens-lazy-loading';
+import { TDocumentDataLazyLoadingParams } from './types';
+
+const createPendingPromise = () => {
```
iSokrat, line 6: This helper is needed to help signalise that all needed data is loaded at some moment (`InfiniteLoader` component requires to return resolved promise in `loadMoreItems` callback when all data is loaded).

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -316,77 +350,87 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
...
+    let documentData;
+
+    // TODO: This is a workaround to use different loading approaches in one big context.
+    // For PDF and other type of doc different context should be used.
+    // This logic should be moved in appropriate places where it's needed (this
+    // requires refactoring).
+    if (!isDocumentExist) {
+        documentData = defaultDocumentData;
+    } else {
+        documentData = isOnePDFLoadingFlow
```
iSokrat, line 415: I've moved the previous logic where we load all data in the separate hook, the same about the new logic. The logic in inside `documentDataViaFullLoading` is just a movement of the code which we had before.`documentDataViaLazyLoading` contains new logic about lazy loading.

web/src/connectors/task-annotator-connector/use-document-data-lazy-loading/use-annotations-lazy-loading.ts
```diff
@@ -0,0 +1,96 @@
...
+                    // unset selected annotation if it's not in cache (we render only items which are in cache)
+                    setSelectedAnnotation((prevSelectedAnnotation) => {
+                        const noSelectedAnnotationInCache =
+                            prevSelectedAnnotation &&
+                            !pageNumbersToKeepInCache.has(prevSelectedAnnotation.pageNum!);
+
+                        return noSelectedAnnotationInCache ? undefined : prevSelectedAnnotation;
+                    });
+
+                    // merge old cached data with the new received data
```
iSokrat, line 69: We merge the old loaded data and new data in the following way: 
Since BE can return us just list of new data at this moment, we can't just join old list and new list. All data from requests from request will be the same except several fields like "pages", "validated" and "failed_validation_pages". After discussion with BE, we agreed that FE will just join these lists.

web/src/connectors/task-annotator-connector/use-document-data-lazy-loading/use-annotations-lazy-loading.ts
```diff
@@ -0,0 +1,96 @@
...
+                    setCachedRange(availableRenderedPagesRange, nextLoadingPagesRange);
+
+                    if (!oldData) {
+                        return newData;
+                    }
+
+                    const pageNumbersToKeepInCache = getPageNumbersToKeepInCache(
+                        pageNumbers,
+                        availableRenderedPagesRange
+                    );
```
iSokrat, line 58: The main idea of this PR is to load the data for PDF file in lazy mode but to not keep all loaded data in cache. In case if we keep all this data we may face a situation that for big documents we may have a lot of cached data which will crash browser's tab. In such situation, we'll only keep some amount of cached data. It means that in case if we loaded some data for some PDF pages then it does NOT mean this data will not be refetched anymore.

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -208,6 +238,7 @@ export const TaskAnnotatorContextProvider: React.FC<ProviderProps> = ({
 
     const [currentPage, setCurrentPage] = useState<number>(1);
 
+    const [annotationsChanges, setAnnotationsChanges] = useState<Record<number, Annotation[]>>({});
```
iSokrat, line 241: In the scope of this PR we start to store all the changes related to annotations in the separate map. Before changes were just directly added to the total list of annotations. It's needed because we don't store all the items in the total list of annotations now when lazy loading mode is enabled for PDF (in lazy loading mode this list is constantly changed since we loading a new data during scroll and remove old cached data). Now, these changes are fully tracked separately.

web/src/shared/components/document-pages/components/document-pdf/index.tsx
```diff
@@ -143,33 +139,41 @@ const DocumentPDF: React.FC<DocumentPDFProps> = ({
 
     // in case of scrolling to some annotation
     useEffect(() => {
-        if (!selectedAnnotation || !pdfPagesListRef.current || !listViewContainerRef.current) {
-            return;
-        }
+        // give the browser chance to make additional painting (to wait for changes with
+        // annotations like creation of new annotation)
+        requestAnimationFrame(() => {
```
iSokrat, line 144: Ideally, should be refactored in the future to avoid "requestAnimationFrame" at all

Ziprion (approved):

Ziprion (approved):

Issue #729, opened 2023-08-30
Document can't be selected


iSokrat:
Document can't be selected - after clicking on the checkbox the page is immediately opened.

<img width="1422" alt="Screenshot 2023-08-30 at 14 21 40" src="https://github.com/epam/badgerdoc/assets/13063442/94f93531-86f1-41e7-b12d-280304821c63">

PR #735, opened 2023-09-12
fix: document cant be selected

Ziprion:
Closes [#729](https://github.com/epam/badgerdoc/issues/729)


web/src/components/documents/document-card-view-item/document-card-view-item.module.scss
```diff
@@ -2,7 +2,7 @@
     position: relative;
```
iSokrat, line 2: Do we really need to show this border by default? Should id be only appear when we drop something there?
image

iSokrat, line 2: Just noticed that the color of ticks was changed to grey (not sure that this PR affected this but maybe some previous)?
image

On dev2 we have white ticks
image

web/src/pages/documents/documents-page.tsx
```diff
@@ -95,14 +95,15 @@ const DocumentsPage = () => {
...
+                                    flexDirection: 'column'
+                                }}
                             >
-                                {!isFileOver && (
-                                    <div
-                                        style={{
-                                            height: '100%'
-                                        }}
-                                    >
+                                <DocumentsDropZone dataset={activeDataset}>
```
iSokrat, line 105: Do you know how should drag & drop work when we move file via "in documents" tab? I've tried to move usual jpg there, it was successfully loaded but shown in "in file names" tab. Should we somehow automatically redirect the user to the place where it was placed? Or it should appear in "in documents" tab, do you know? Is drag & drop actually needed for "in documents" tab or not?

Ziprion: Actually, I don't know how it should work. Anyway, it's out of scope of this PR

web/src/components/documents/document-card-view-item/document-card-view-item.module.scss
Ziprion: I think, we should show that it's a dnd zone and user could use this feature. In additional, we have different styles for dnd zone during overing files(blue borders and background)

iSokrat: Okay,  let's keep it

web/src/pages/documents/documents-page.tsx
iSokrat: Okay, moved this to "Open questions"

web/src/components/documents/document-card-view-item/document-card-view-item.module.scss
Ziprion: fixed

iSokrat (approved):

PR #738, opened 2023-09-15
fix: Fixes Finish Validation/Labelling Button Tooltip

ammb95:
In the scope of this PR, the conditions to show a tooltip for the FinishButton are fixed

### DIFF
--- a/web/src/components/task/task-sidebar/finish-button/finish-button.tsx
+++ b/web/src/components/task/task-sidebar/finish-button/finish-button.tsx
@@ -52,7 +52,7 @@ export const FinishButton: FC<TaskSidebarProps> = ({
 
     const handleFinishValidation = isSplitValidation ? onFinishSplitValidation : onFinishValidation;
 
-    let isDisabled = !isAnnotatable;
+    let isDisabled = !isAnnotatable || (isValidation && !allValidated);
     if (jobType !== 'extensive_coverage') {
         isDisabled =
             !isAnnotatable ||
@@ -60,10 +60,9 @@ export const FinishButton: FC<TaskSidebarProps> = ({
             (isValidation && !touchedPagesCount && !editedPagesCount);
     }
 
-    const tooltipContent =
-        isDisabled && allValidated && !isSplitValidation
-            ? ''
-            : createTooltip(jobType === 'extensive_coverage', notProcessedPages);
+    const tooltipContent = !isDisabled
+        ? null
+        : createTooltip(jobType === 'extensive_coverage', notProcessedPages);
 
     const redirectSettingsDataSource = useArrayDataSource(
         {


web/src/components/task/task-sidebar/finish-button/finish-button.tsx
```diff
@@ -52,18 +52,23 @@ export const FinishButton: FC<TaskSidebarProps> = ({
...
     const handleFinishValidation = isSplitValidation ? onFinishSplitValidation : onFinishValidation;
 
-    let isDisabled = !isAnnotatable;
-    if (jobType !== 'extensive_coverage') {
-        isDisabled =
-            !isAnnotatable ||
-            !allValidated ||
-            (isValidation && !touchedPagesCount && !editedPagesCount);
-    }
+    const isDisabled = useMemo(() => {
```
iSokrat, line 55: Which reason is to use `useMemo` here? The logic is not resource-intensive here, we can just create a variable and save there needed condition

web/src/components/task/task-sidebar/finish-button/finish-button.tsx
```diff
@@ -52,18 +52,23 @@ export const FinishButton: FC<TaskSidebarProps> = ({
...
+            );
+        }
+        return !isAnnotatable || (isValidation && !allValidated);
+    }, [jobType, isAnnotatable, allValidated, isValidation, touchedPagesCount, editedPagesCount]);
 
-    const tooltipContent =
-        isDisabled && allValidated && !isSplitValidation
-            ? ''
-            : createTooltip(jobType === 'extensive_coverage', notProcessedPages);
+    const tooltipContent = useMemo(() => {
```
iSokrat, line 66: The same question as above

web/src/components/task/task-sidebar/finish-button/finish-button.tsx
ammb95: True, addresed

web/src/components/task/task-sidebar/finish-button/finish-button.tsx
ammb95: Addressed

web/src/components/task/task-sidebar/finish-button/finish-button.tsx
```diff
@@ -52,18 +52,17 @@ export const FinishButton: FC<TaskSidebarProps> = ({
...
+    let isDisabled = !isAnnotatable || (isValidation && !allValidated);
     if (jobType !== 'extensive_coverage') {
         isDisabled =
             !isAnnotatable ||
             !allValidated ||
             (isValidation && !touchedPagesCount && !editedPagesCount);
     }
 
-    const tooltipContent =
-        isDisabled && allValidated && !isSplitValidation
```
iSokrat, line 64: I see you've removed "&& allValidated && !isSplitValidation", before it was also combined with the "isDisabled" placed under "if". Is it needed or this logic should still be presented for the logic under "jobType !== 'extensive_coverage'"?

ammb95: I rearranged the logic so that, for non extensive coverage cases, `isDisabled` is true in the case `(isValidation && !allValidated)` is `true`. (I.e., the button is disabled if we are under validation and not all pages are validated). So now, `isDisabled` accounts for `allValidated`

From my analysis of the code, it is logical the the tooltip should only appear In the case the button is disabled. Keeping the `isSplitValidation` on the `tootipContent` calculation will make the tooltip appear in some cases the button is not disabled.

At last, the logic inside the `if` statement was not changed.

web/src/components/task/task-sidebar/finish-button/finish-button.tsx
```diff
@@ -52,18 +52,17 @@ export const FinishButton: FC<TaskSidebarProps> = ({
```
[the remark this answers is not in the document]
iSokrat: Hm, for me now it's also disabled 🤔

iSokrat (approved):

PR #739, opened 2023-09-19
fix: Tooltip With Very Long List Of Pages

ammb95:
This PR addresses the issue of a very long list of remaining pages appearing on the tooltip of the finish validation button.

It consists on the creation of a set of utils that will handle the cases when more than 10 pages are remaining

### DIFF
--- a/web/src/components/task/task-sidebar/finish-button/utils.ts
+++ b/web/src/components/task/task-sidebar/finish-button/utils.ts
@@ -1,14 +1,12 @@
 const LISTED_PAGES_LIMIT = 10;
-const SLICE_START_INDEX = 0;
-const SLICE_END_INDEX = LISTED_PAGES_LIMIT - 1;
 
-const getPagesList = (pages: number[]) => pages?.join(', ');
+const getPagesList = (pages: number[]) => pages.join(', ');
 
 const getAboveLimitPagesCount = (pages: number[]) => pages.length - LISTED_PAGES_LIMIT;
 
-const getVisiblePagesString = (pages: number[]) => getPagesList(getVisiblePagesArray(pages));
+const getVisiblePagesArray = (pages: number[]) => pages.slice(0, LISTED_PAGES_LIMIT);
 
-const getVisiblePagesArray = (pages: number[]) => pages.slice(SLICE_START_INDEX, SLICE_END_INDEX);
+const getVisiblePagesString = (pages: number[]) => getPagesList(getVisiblePagesArray(pages));
 
 const listRemainingPages = (pages: number[]) => {
     if (pages.length > LISTED_PAGES_LIMIT) {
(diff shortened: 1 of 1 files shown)


web/src/components/task/task-sidebar/finish-button/utils.ts
```diff
@@ -1,7 +1,29 @@
...
+const SLICE_START_INDEX = 0;
+const SLICE_END_INDEX = LISTED_PAGES_LIMIT - 1;
+
+const getPagesList = (pages: number[]) => pages?.join(', ');
+
+const getAboveLimitPagesCount = (pages: number[]) => pages.length - LISTED_PAGES_LIMIT;
+
+const getVisiblePagesString = (pages: number[]) => getPagesList(getVisiblePagesArray(pages));
+
+const getVisiblePagesArray = (pages: number[]) => pages.slice(SLICE_START_INDEX, SLICE_END_INDEX);
```
iSokrat, line 11: Let's move this before "getVisiblePagesString" since in the future we may accidentally catch the case when "getVisiblePagesString" is called earlier than "getVisiblePagesArray" is defined.

web/src/components/task/task-sidebar/finish-button/utils.ts
```diff
@@ -1,7 +1,29 @@
+const LISTED_PAGES_LIMIT = 10;
+const SLICE_START_INDEX = 0;
+const SLICE_END_INDEX = LISTED_PAGES_LIMIT - 1;
+
+const getPagesList = (pages: number[]) => pages?.join(', ');
```
iSokrat, line 5: Do we really need to use "?" here. Based on "number[]" type, it's always defined

web/src/components/task/task-sidebar/finish-button/utils.ts
iSokrat, line 11: Optional: I think we can just directly use numbers (based on LISTED_PAGES_LIMIT) instead of "SLICE_START_INDEX" and "SLICE_END_INDEX". This will not look like magic numbers

web/src/components/task/task-sidebar/finish-button/utils.ts
```diff
@@ -1,7 +1,29 @@
+const LISTED_PAGES_LIMIT = 10;
+const SLICE_START_INDEX = 0;
+const SLICE_END_INDEX = LISTED_PAGES_LIMIT - 1;
```
iSokrat, line 3: Are you sure we need this "-1" here? As I understand, in this case we'll show 9 numbers instead of 10 and 1 page number will be lost when we do "pages.length - LISTED_PAGES_LIMIT".

web/src/components/task/task-sidebar/finish-button/utils.ts
ammb95: Addressed

web/src/components/task/task-sidebar/finish-button/utils.ts
ammb95: Addressed

web/src/components/task/task-sidebar/finish-button/utils.ts
ammb95: Addressed

web/src/components/task/task-sidebar/finish-button/utils.ts
ammb95: Addressed

iSokrat (approved):
@ammb95 Can you please check this PR after deployment?

PR #744, opened 2023-09-22
test: update tests for "AnnotationList" component

iSokrat:
In the scope of this PR, tests for "AnnotationList" component were fixed.

### DIFF
--- a/web/src/connectors/task-annotator-connector/task-annotator-context.tsx
+++ b/web/src/connectors/task-annotator-connector/task-annotator-context.tsx
@@ -82,7 +82,7 @@ import { useDocumentDataFullLoading } from './use-document-data-full-loading';
 
 type WithFunctionStub<T> = T | (() => void);
 
-type ContextValue = SplitValidationValue &
+export type ContextValue = SplitValidationValue &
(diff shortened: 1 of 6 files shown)


web/src/connectors/task-annotator-connector/task-annotator-context.tsx
```diff
@@ -82,7 +82,7 @@ import { useDocumentDataFullLoading } from './use-document-data-full-loading';
 
 type WithFunctionStub<T> = T | (() => void);
 
-type ContextValue = SplitValidationValue &
+export type ContextValue = SplitValidationValue &
```
ammb95, line 85: Do you think a more specific name to this interface would go well?

ammb95 (approved):
The comment I left is a minor thing and regardless that, approved

web/src/connectors/task-annotator-connector/task-annotator-context.tsx
iSokrat: I think, yes. Just didn't rename it to not touch this file a lot in this PR. We can update it in the following PRs

PR #748, opened 2023-09-26
fix: fix showing of valid/invalid buttons

iSokrat:
Steps to reproduce:

1. Go to BadgerDoc platform
2. Open a validation task with "In progress" status
3. Mark any page as "Invalid"
4. Make some annotation changes
5. Press "Save edits"

Expected result:

The changes are saved.
User is able to continue validation the page.

Actual result:

Validation buttons (valid/invalid) disappear.
User is not able to validate the page.


web/src/components/task/task-sidebar/task-sidebar.tsx
```diff
@@ -512,70 +512,72 @@ const TaskSidebar: FC<TaskSidebarProps> = ({ jobSettings, viewMode, isNextTaskPr
                             {isValidation && !splitValidation && (
```
ammb95, line 512: What do you think about moving all this wrapper together with the buttons to a separated component?

web/src/components/task/task-sidebar/finish-button/finish-button.tsx
```diff
@@ -53,17 +49,11 @@ export const FinishButton: FC<TaskSidebarProps> = ({
 
     const handleFinishValidation = isSplitValidation ? onFinishSplitValidation : onFinishValidation;
 
-    let isDisabled = !isAnnotatable || (isValidation && !allValidated);
-    if (jobType !== 'extensive_coverage') {
```
ammb95, line 57: Why was the if statement removed?

ammb95 (approved):

web/src/components/task/task-sidebar/task-sidebar.tsx
iSokrat: I was thinking about this but decided to leave as is for now - it's + - okay at this stage

web/src/components/task/task-sidebar/finish-button/finish-button.tsx
iSokrat: Looks like it's not needed and it's okay to use 1 logic for all cases. `(isValidation && !touchedPagesCount && !editedPagesCount);` part counted the case incorrectly - in case if all pages are validated and not touched anyhow, it should be possible to click "Finish validation" but this check didn't allow to do this. I've tried to figured out when it was introduced (and the reason of this check) and haven't found any reason to keep it here.

PR #753, opened 2023-09-27
fix: Edit Categories Modal

ammb95:
This PR addresses the issues described of the edit categories modal

For the given context of editing an existing category:
- Adds a new class to hide category-id field
- Fixes utils the map existing category values to the form values
- Implements necessary changes on Modal component

### DIFF
--- a/web/src/connectors/categories-modal-connector/categories-modal-connector.tsx
+++ b/web/src/connectors/categories-modal-connector/categories-modal-connector.tsx
@@ -107,11 +107,14 @@ export const ModalWithDisabledClickOutsideAndCross: FC<IProps> = ({
         value: getDefaultValues(categoryValue)
     });
 
+    const title = categoryValue ? 'Edit category' : 'Add new category';
+    const categoryIdClassName = categoryValue ? styles['hidden-category-id'] : '';
+
     return (
         <ModalBlocker disallowClickOutside blockerShadow="dark" {...props} abort={onClose}>
             <ModalWindow>
                 <Panel background="white">
-                    <ModalHeader title="Add new category" />
+                    <ModalHeader title={title} />
                     <ScrollBars hasTopShadow hasBottomShadow>
                         <FlexRow padding="24" vPadding="12">
                             <FlexCell grow={1}>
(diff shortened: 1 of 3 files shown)


web/src/connectors/categories-modal-connector/utils.ts
```diff
@@ -39,11 +40,11 @@ export const getDefaultValues = (category: Category | undefined) => {
     } = category || {};
 
     return {
-        categoryId: '',
+        categoryId: id || '',
         name: name ?? '',
         parentId: parent,
         type: type ?? 'box',
-        withParentLabel: !parent,
+        withParentLabel: !!parent,
```
iSokrat, line 47: Why did we update this condition here?

ammb95: The way it was:
- When creating a category, parent checkbox was automatically selected by default. So, if you didnt want it to have a parent, you needed to manually uncheck the checkbox.
- When updating a category without parent, it was automatically selected to have a parent as well, so you needed to manually uncheck the checkbox

This way
- On category creation, checkbox starts unselected
- When updating a category that does not have parent, checkbox starts unchecked
- When updating a category that has a parent, checkbox comes checked and the previously selected parent comes selected

web/src/connectors/categories-modal-connector/categories-modal-connector.tsx
```diff
@@ -120,13 +123,15 @@ export const ModalWithDisabledClickOutsideAndCross: FC<IProps> = ({
...
                             </FlexCell>
                         </FlexRow>
-                        <FlexRow padding="24" vPadding="12">
-                            <FlexCell grow={1}>
-                                <LabeledInput isRequired label="Category Id">
-                                    <TextInput {...lens.prop('categoryId').toProps()} />
-                                </LabeledInput>
-                            </FlexCell>
-                        </FlexRow>
+                        <div className={categoryIdClassName}>
```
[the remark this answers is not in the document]
ammb95: I tried it but, as the field will not exist in this case, there will be a problem with the form, this way, the field is hidden as the AC asks but the form still gets the value from the field

iSokrat (approved):

PR #755, opened 2023-09-27
fix: show "confirm" modal only when annotation is finished

Nosafi:
Confirm modal only when annotation finished. 
New text added to confirm modal.


iSokrat:
Let's name the PR more correctly: `fix: show "confirm" modal only when annotation is finished`

iSokrat (approved):

PR #756, opened 2023-09-28
chore: Improve code checkers for "web" folder

iSokrat:
In the scope of this PR, several changes are introduced:
1) `.eslintrc.js` is updated: 
  1.1) `eslint-plugin-prettier` is removed in order to more follow "opionated" rules of "prettier" (this plugin is not generally recommended by prettier tool);
  1.2) `create-react-app`'s plugins are added.
2) `package.json` - some linting/formatting-related scripts are re-organised, `--fix` option for `eslint` is removed to only validate the rules without introducing automatic changes (this can be done seprately), latest ecm script is defined
3) `check:format` script is added to pre-commit in order to validate formatting-related stuff.
4) after updating of the rules for eslint, a lot of issues were found in the codebase ("/web"). All of these issues were **temporary** disabled and should be fixed separately. Disabling was done in most cases automatically (via `eslint-interactive` package).
5) small prettier-related issue is fixed.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,3 +1,5 @@
+// temporary_disabled_rules
+/* eslint-disable @typescript-eslint/no-unused-vars, react-hooks/exhaustive-deps */
 import React, { useEffect, useState } from 'react';
 import { Redirect, Route, Switch, useHistory } from 'react-router-dom';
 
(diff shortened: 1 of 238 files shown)


web/src/App.tsx
```diff
@@ -1,3 +1,5 @@
+// temporary_disabled_rules
+/* eslint-disable @typescript-eslint/no-unused-vars, react-hooks/exhaustive-deps */
 import React, { useEffect, useState } from 'react';
```
ammb95, line 3: Why don't we remove all unused occurrences of React?

ammb95 (approved):

web/src/App.tsx
iSokrat: It's planned to be done in the separate PR.

PR #759, opened 2023-10-02
fix: introduces prevToken evaluation into new line calculation

ammb95:
This PR fixes the behavior of creating a text annotation.

The previous code was not identifying line breaks correctly, so a new condition for line break was added.

### DIFF
--- a/web/src/shared/components/annotator/components/text-annotation/helpers.ts
+++ b/web/src/shared/components/annotator/components/text-annotation/helpers.ts
@@ -89,8 +89,18 @@ export const updateCurrentTextAnnBound = (curBound: Bound, token: PageToken): Bo
     };
 };
 
-export const isTokenOnNewLine = (token: PageToken, curBound: Bound): boolean => {
-    return isBoundValid(curBound) && token.y > curBound.y + token.height;
+const LINE_BREAK_CHAR = '\n';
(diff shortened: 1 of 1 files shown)


web/src/shared/components/annotator/components/text-annotation/helpers.ts
```diff
@@ -89,8 +89,18 @@ export const updateCurrentTextAnnBound = (curBound: Bound, token: PageToken): Bo
     };
 };
 
-export const isTokenOnNewLine = (token: PageToken, curBound: Bound): boolean => {
-    return isBoundValid(curBound) && token.y > curBound.y + token.height;
+const LINE_BREAK_CHAR = '\n';
+const isTokenLastOfLine = (token: PageToken) => token?.text!.includes(LINE_BREAK_CHAR);
```
iSokrat, line 93: Is "?" and "!" needed here? As I see based on the type should be defined here or there are some cases when something goes wrong?

ammb95: When prevToken is declared, it is declared as an empty object and casted to Bound. During the first check, it could cause problems so I used both.

iSokrat (approved):

PR #760, opened 2023-10-02
fix: added 'Validation only' desc in validation picker

Nosafi:
Added 'Validation only' description in validation picker


ammb95 (approved):
One suggestion of improvement is moving everything inside the description prop to a separate component. Other than that, approved

PR #763, opened 2023-10-03
fix: Close Create Task Modal On Save

ammb95:
This PR aims to add logic to close the CreateTask Modal on successful task creation.

The approach consists on passing down a handleCloseModal function as a prop to the CreateTask component

### DIFF
--- a/web/src/connectors/tasks/create-task/create-task.tsx
+++ b/web/src/connectors/tasks/create-task/create-task.tsx
@@ -27,12 +27,14 @@ interface ICreateNewTaskProps extends IModal<object> {
     annotatorIds: Array<string>;
     fileIds: Array<number>;
     jobId?: number;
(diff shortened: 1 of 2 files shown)


web/src/connectors/tasks/create-task/create-task.tsx
```diff
@@ -73,6 +74,7 @@ export const CreateTask: FC<ICreateNewTaskProps> = ({
                         <Text>{`Task is created with id=${response.id}`}</Text>
                     </>
                 );
+                svc.uuiModals.closeAll();
```
iSokrat, line 77: I propose to close the modal outside "create-task.tsx"

ammb95: Logic changed

iSokrat (approved):

PR #764, opened 2023-10-03
feat: reflect currently shown page's order number in "Current page" dropdown when page is scrolled

iSokrat:
In the scope of this PR, currently shown page's order number is in the dropdown when page is scrolled. "Go to page" is renamed to "Current page".

### DIFF
--- a/web/src/pages/document/document-page.tsx
+++ b/web/src/pages/document/document-page.tsx
@@ -1,6 +1,6 @@
 // temporary_disabled_rules
 /* eslint-disable react-hooks/exhaustive-deps */
-import React, { useCallback, useState } from 'react';
+import React, { useState } from 'react';
 
 import {
     DocumentPageSidebarContent,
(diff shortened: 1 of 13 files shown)


web/src/pages/task/task-page.tsx
```diff
@@ -19,53 +19,110 @@ import { FlowSideBar } from 'components/task/task-sidebar-flow/task-sidebar-flow
...
+
+type TTaskPageContent = {
+    taskId: string;
+    handleRedirectToNextTask: () => void;
+    history: History<Record<string, string | undefined>>;
+    nextTaskId: number | undefined;
+};
 
-const TaskPage: FC = () => {
+const TaskPageContent: FC<TTaskPageContent> = ({
```
ammb95, line 35: One suggestion for improvement would be to move this component to a separate file

web/src/pages/document/document-page.tsx
```diff
@@ -70,97 +68,42 @@ export function DocumentPage({
...
-                                    cx={styles.button}
-                                    onClick={handleGoNext}
-                                    isDisabled={isLastPage}
-                                />
-                            </FlexRow>
+        <TaskAnnotatorContextProvider
+            jobId={documentJobId}
+            revisionId={documentJobRevisionsInfo?.selectedDocumentJobRevisionId}
+            fileMetaInfo={fileMetaInfo}
+            onRedirectAfterFinish={() => {}}
```
ammb95, line 82: For convention reasons, what do you think about using that `noop` function instead of this empty callback?

ammb95 (approved):

web/src/pages/document/document-page.tsx
iSokrat: This logic was not introduced by me, this was just move from another place. I think "noop" is actually not a convention for the project, this was just someone's wish + we have a task to remove "lodash" from the project at all, so, I think it's okay for now to leave it as is.

web/src/pages/task/task-page.tsx
iSokrat: We can do this as an improvement in the future

PR #771, opened 2023-10-09
fix: blocking deletion of annotations in some scenarios.

Nosafi:
If the task is in the finished status, and also during validation outside of page editing, the delete of annotations function becomes impossible.

### DIFF
--- a/web/src/components/task/task-sidebar-flow/annotation-list.tsx
+++ b/web/src/components/task/task-sidebar-flow/annotation-list.tsx
@@ -93,7 +93,7 @@ export const AnnotationList: FC<AnnotationListProps> = ({
     const isClosable =
         task?.status !== 'Finished'
             ? task?.is_validation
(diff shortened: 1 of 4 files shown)


web/src/components/task/task-sidebar-flow/annotation-list.tsx
```diff
@@ -76,13 +76,27 @@ export const AnnotationList: FC<AnnotationListProps> = ({
...
+    } = useTaskAnnotatorContext();
     const { incomingLinksByAnnotationId, annotationNameById } = useMemo(
         () => collectIncomingLinks(list),
         [list]
     );
 
+    const isClosable =
+        task?.status !== 'Finished'
+            ? task?.is_validation
+                ? editedPages.includes(currentPage) !== false
```
iSokrat, line 96: Can be just `editedPages.includes(currentPage)`

PR #772, opened 2023-10-10
feat: Post upload processing notifications

ammb95:
The goal of this PR is to show notifications when file processing takes more than 10 seconds.

After first notification, next one will be in 20sec and on every time notification is shown, delay is doubled.

For that was implemented `useNotifyProcessing` hook, which will watch `isUploadedValue` from `useFetchWithTrackingOfUploadProgress` and trigger the notifications based on that.

### DIFF
--- a/web/src/connectors/documents-page-control-connector/fetch-with-tracking-of-upload-progress-factory.ts
+++ b/web/src/connectors/documents-page-control-connector/fetch-with-tracking-of-upload-progress-factory.ts
@@ -6,21 +6,23 @@ const calculateProgress = (loaded: number, total: number): number => {
     return Math.floor(100 * (loaded / total));
 };
 
-const handleUploadFailed = (xhr: XMLHttpRequest) => {
+const handleUploadFailed = (xhr: XMLHttpRequest, externalCallback: () => void) => {
+    externalCallback();
     throw new ApiError(xhr.statusText, `Upload failed with status ${xhr.status}`, {
         status: xhr.status,
         ...xhr.response.body
     });
 };
 
-type WithOnProgressCallback = {
+type CustomFetchFactoryDeps = {
     onProgressCallback: UploadProgressTracker['setProgress'];
+    onError: () => void;
 };
 
-export type TFetchType = (args: WithOnProgressCallback) => BadgerCustomFetch;
+export type TFetchType = (deps: CustomFetchFactoryDeps) => BadgerCustomFetch;
 
 export const fetchWithTrackingOfUploadProgressFactory: TFetchType =
-    ({ onProgressCallback }) =>
+    ({ onProgressCallback, onError }) =>
     (url, { method, body, headers }) => {
         return new Promise<Response>((resolve) => {
             const xhr = new XMLHttpRequest();
@@ -45,12 +47,12 @@ export const fetchWithTrackingOfUploadProgressFactory: TFetchType =
                         })
                     );
                 } else {
-                    handleUploadFailed(xhr);
+                    handleUploadFailed(xhr, onError);
                 }
             };
 
             xhr.onerror = () => {
-                handleUploadFailed(xhr);
+                handleUploadFailed(xhr, onError);
             };
 
             xhr.send(body as FormData);
--- a/web/src/connectors/documents-page-control-connector/fetch-with-tracking-of-upload-progress-factory.ts
+++ b/web/src/connectors/documents-page-control-connector/fetch-with-tracking-of-upload-progress-factory.ts
@@ -6,14 +6,6 @@ const calculateProgress = (loaded: number, total: number): number => {
     return Math.floor(100 * (loaded / total));
 };
 
-const handleUploadFailed = (xhr: XMLHttpRequest, externalCallback: () => void) => {
-    externalCallback();
-    throw new ApiError(xhr.statusText, `Upload failed with status ${xhr.status}`, {
-        status: xhr.status,
-        ...xhr.response.body
-    });
-};
-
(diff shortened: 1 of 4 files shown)


web/src/connectors/documents-page-control-connector/use-notify-processing.ts
```diff
@@ -0,0 +1,43 @@
...
+import { NotificationTypes, showNotification } from 'shared/components/notifications';
+
+const notifyProcessing = () => {
+    showNotification({
+        type: NotificationTypes.hint,
+        text: 'Your file is still processing. This may take a while'
+    });
+};
+
+export const useNotifyProcessing = (isUploaded: boolean) => {
```
iSokrat, line 11: In case of failed request - will this still work?

web/src/connectors/documents-page-control-connector/use-notify-processing.ts
```diff
@@ -0,0 +1,43 @@
...
+    useEffect(() => {
+        let delay = 10000;
+        let timeoutId: number | null = null;
+        let firstNotificationShown = false;
+
+        const showNextNotification = () => {
+            if (isUploaded) {
```
iSokrat, line 18: In order to reduce nesting, we can use before the main logic "if (!isUploaded) {return;}"

web/src/connectors/documents-page-control-connector/use-notify-processing.ts
```diff
@@ -0,0 +1,43 @@
...
+                if (!firstNotificationShown) {
+                    timeoutId = setTimeout(() => {
+                        firstNotificationShown = true;
+                        showNextNotification();
+                    }, delay) as unknown as number;
```
iSokrat, line 23: Instead of doing this casting, I propose to use "ReturnType<typeof setTimeout> | null" instead of "number | null" for "timeoutId"

web/src/connectors/documents-page-control-connector/use-notify-processing.ts
```diff
@@ -0,0 +1,43 @@
...
+        let delay = 10000;
+        let timeoutId: number | null = null;
+        let firstNotificationShown = false;
+
+        const showNextNotification = () => {
+            if (isUploaded) {
+                if (!firstNotificationShown) {
+                    timeoutId = setTimeout(() => {
+                        firstNotificationShown = true;
+                        showNextNotification();
```
iSokrat, line 22: Am I right that after 10 secs we don't call "notifyProcessing"?

ammb95: No, on the first round, `firstNotificationShown` is false, so the first timeout is set calling `showNextNotification` on the timeout callback. and setting `firstNotificationShown` as true

after 10 secs, `showNextNotification` is called again because of the last timeout, but this time `firstNotificationShown` is true (it was set as true 10 secs before this round), and we fall into the else block which immediately calls the `notifyProcessing` function and sets a new timeout with double delay

web/src/connectors/documents-page-control-connector/use-notify-processing.ts
ammb95: addressed

web/src/connectors/documents-page-control-connector/use-notify-processing.ts
ammb95: The notification of request error is done separately from this hook, please check new changes

web/src/connectors/documents-page-control-connector/use-notify-processing.ts
iSokrat: > No, on the first round, firstNotificationShown is false, so the first timeout is set calling showNextNotification on the timeout callback. and setting firstNotificationShown as true

But why on the first round we don't show notification?

web/src/connectors/documents-page-control-connector/use-fetch-with-tracking-of-upload-progress.ts
```diff
@@ -10,7 +12,23 @@ export interface UploadProgressTracker {
...
-export const useFetchWithTrackingOfUploadProgress = (): UploadProgressTracker => {
+type UseFetchWithTrackingOfUploadProgressDeps = {
+    onError: () => void;
+};
+
+const useHandleUploadError = (onError: () => void) => {
+    return useCallback(() => {
+        showNotification({
+            type: NotificationTypes.error,
+            text: 'Error uploading the file. Try again later.'
```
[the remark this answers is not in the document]
ammb95: addressed

web/src/connectors/documents-page-control-connector/fetch-with-tracking-of-upload-progress-factory.ts
```diff
@@ -6,21 +6,23 @@ const calculateProgress = (loaded: number, total: number): number => {
     return Math.floor(100 * (loaded / total));
 };
 
-const handleUploadFailed = (xhr: XMLHttpRequest) => {
+const handleUploadFailed = (xhr: XMLHttpRequest, externalCallback: () => void) => {
+    externalCallback();
```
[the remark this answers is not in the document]
ammb95: addressed

web/src/connectors/documents-page-control-connector/use-fetch-with-tracking-of-upload-progress.ts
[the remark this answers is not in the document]
iSokrat: UPD: decided to not use it at all since there is already another notification appeared in such case.

iSokrat (approved):

PR #813, opened 2023-11-29
fix: fixes in Readme

khyurri:


clearml/docker-compose-dev.yaml
```diff
@@ -0,0 +1,211 @@
...
+    - clearml-data-fileserver:/mnt/fileserver
+    depends_on:
+      - redis
+      - mongo
+      - elasticsearch
+      - fileserver
+    environment:
+      CLEARML_ELASTIC_SERVICE_HOST: elasticsearch
+      CLEARML_ELASTIC_SERVICE_PORT: 9200
+      CLEARML_ELASTIC_SERVICE_PASSWORD: ${ELASTIC_PASSWORD}
```
cakeinsauce, line 22: Maybe move it to secrets? If it's acceptable to keep here, then okay

khyurri: We're using docker-compose for dev installations only where we don't use any secrets manager. For k8s installation these variables must be populated by secrets.
However, exact this docker-compose-dev.yaml doesn't specify how secretes are populated. I assume easiest way to do it - create `.env` file inside `clearml` directory

cakeinsauce (approved):
LGTM!

PR #831, opened 2024-04-23
feat: add sdf pagination support

khyurri:


assets/assets/exceptions.py
```diff
@@ -28,3 +28,11 @@ class FileConversionError(Exception):
     """
 
     pass
+
+
+class AssetsUnsupportedFileFormat(Exception):
+    pass
+
+
+class AssetsMaxMoleculesCountExceeded(Exception):
```
cakeinsauce, line 37: Should we specify `Assets` at the beginning of the exception name? It is located in assets microservice anyway

web/src/shared/helpers/bbox-to-bound.ts
```diff
@@ -1,8 +1,18 @@
...
+            height: 0
+        }
+    }
+    return {
+        x: bbox[0] * scale,
+        y: bbox[1] * scale,
+        width: (bbox[2] - bbox[0]) * scale,
+        height: (bbox[3] - bbox[1]) * scale
+    }
+}
```
cakeinsauce, line 18: Could you add new line at the end o a file?

assets/assets/utils/chem_utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+SUPPORTED_FORMATS = {
+    ".pdf": "pdf",
+    ".mol": "chem",
+    ".sdf": "chem",
+    ".png": "image",
+    ".jpg": "image",
+    ".jpeg": "image",
+    ".txt": "text",
+}
```
cakeinsauce, line 17: Could we make it, let's say, enum or named tuple? This way we won't compare hardcoded extensions in the code

assets/assets/utils/chem_utils.py
```diff
@@ -0,0 +1,60 @@
...
+MAX_MOLECULES_COUNT_IN_FILE = os.getenv("MAX_MOLECULES_COUNT_IN_FILE", 50)
+
+
+def get_page_count(file: bytes, ext: str) -> int:
+    indigo = Indigo()
+    pages: int = 1
+    if ext == ".sdf":
+        for pages, _ in enumerate(
+            indigo.iterateSDF(indigo.loadBuffer(file)), start=1
+        ):
```
cakeinsauce, line 30: Maybe use range(len()) instead?

assets/assets/utils/chem_utils.py
```diff
@@ -0,0 +1,60 @@
...
+                    MAX_MOLECULES_COUNT_IN_FILE,
+                )
+        logger_.debug("Count of elements in file: %s", pages)
+        return pages
+    return pages
+
+
+def __load_molecule(file: bytes, indigo: Indigo, ext: str) -> Indigo:
+    # in case of .sdf (maybe others types) get only first compound for preview
+    if ext == ".sdf":
```
cakeinsauce, line 44: Please check comment above

assets/assets/utils/chem_utils.py
```diff
@@ -0,0 +1,60 @@
...
+        for mol in indigo.iterateSDF(indigo.loadBuffer(file)):
+            return mol  # getting first element to preview
+    return indigo.loadMolecule(file.decode(encoding="utf-8"))
+
+
+def make_thumbnail(file: bytes, ext: str) -> Union[bool, bytes]:
+    logger_.debug("Generating thumbnail from molecules")
```
cakeinsauce, line 51: Can we add any data to this log? Like file name, or extension?

assets/assets/utils/common_utils.py
```diff
@@ -58,15 +63,22 @@ def get_mimetype(file: bytes) -> Any:
...
-    pages = get_pages_from_pdf(file) or get_pages_from_image(file)
-    return pages
+def get_pages(file: bytes, ext: str) -> Any:
+    logger_.info("Getting count of pages by file extension: %s", ext)
+    if ext not in chem_utils.SUPPORTED_FORMATS:
+        raise exceptions.AssetsUnsupportedFileFormat(
+            "File extension %s is not supported", ext
+        )
+    if "chem" == chem_utils.SUPPORTED_FORMATS[ext]:
+        return chem_utils.get_page_count(file, ext)
```
cakeinsauce, line 73: Please check comment above

assets/assets/utils/chem_utils.py
cakeinsauce: and values of the enum could be tuples or other named tuple with fields `extension` and `category`, where category is another enum.

assets/assets/utils/common_utils.py
```diff
@@ -58,15 +63,22 @@ def get_mimetype(file: bytes) -> Any:
...
+    # Pdf as default
+    return get_pages_from_pdf(file) or get_pages_from_image(file)
 
 
 def get_pages_from_pdf(file: bytes) -> Any:
     try:
         pages = pdf2image.pdfinfo_from_bytes(file)["Pages"]
-    except:  # noqa
+    except Exception:  # noqa
```
cakeinsauce, line 81: Do we need `# noqa` after clarification of exception? Flake8 shouldn't complain about this anymore, but need to check

assets/assets/utils/common_utils.py
```diff
@@ -75,7 +87,7 @@ def get_pages_from_image(file: bytes) -> Any:
     try:
         with PIL.Image.open(BytesIO(file)) as image:
             pages = image.n_frames
-    except:  # noqa
+    except Exception:  # noqa
```
cakeinsauce, line 90: Please check comment above

assets/assets/utils/minio_utils.py
```diff
@@ -70,37 +74,51 @@ def upload_in_minio(
...
-    elif ext == ".pdf":
+
+    if ext not in chem_utils.SUPPORTED_FORMATS:
+        raise exceptions.AssetsUnsupportedFileFormat(
+            "File extension %s is not supported", ext
+        )
+
+    if "chem" == chem_utils.SUPPORTED_FORMATS[ext]:
+        file_bytes = chem_utils.make_thumbnail(obj.data, ext)
+    elif "pdf" == chem_utils.SUPPORTED_FORMATS[ext]:
```
cakeinsauce, line 107: Please check comment above

assets/assets/utils/minio_utils.py
```diff
@@ -70,37 +74,51 @@ def upload_in_minio(
...
         file_bytes = make_thumbnail_pdf(obj.data)
     else:
         logger_.error("Unable to create thumbnail, unsupported extension")
```
cakeinsauce, line 110: Since we are here, could we specify extension is fed into the function?

processing/processing/main.py
```diff
@@ -92,11 +92,18 @@ def get_preprocessing_result(
...
-    )
+    try:
+        bucket_name = convert_bucket_name_if_s3prefix(current_tenant)
+        return Response(
+            content=send_preprocess_result(bucket_name, file_id, pages),
+            media_type="application/json",
+        )
+    except Exception:
+        logger.exception("Cannot fin preprocessing results")
+        return Response(
```
cakeinsauce, line 103: This would work, but you can also use `JsonResponse(content=[])` to make it cleaner

processing/processing/main.py
```diff
@@ -92,11 +92,18 @@ def get_preprocessing_result(
...
-        media_type="application/json",
-    )
+    try:
+        bucket_name = convert_bucket_name_if_s3prefix(current_tenant)
+        return Response(
+            content=send_preprocess_result(bucket_name, file_id, pages),
+            media_type="application/json",
+        )
+    except Exception:
+        logger.exception("Cannot fin preprocessing results")
```
cakeinsauce, line 102:
```suggestion
        logger.exception("Cannot find preprocessing results")
```

assets/assets/utils/chem_utils.py
khyurri: Can you please provide an example, how do you see this mapping can be done using NamedTuple?

assets/assets/utils/chem_utils.py
khyurri: Not sure I got your suggestion. Currently, we're iterating through sdf file until limit is not reached. How can we calculate count of items in sdf without reading it?

processing/processing/main.py
khyurri: Just want to use existing code style. To change `Response` to `JsonResponse` will lead us to change `send_preprocess_result` response. 
In case we will add `JsonResponse` with `Response`, we need to change `get_preprocessing_result` method signature to return `Union[Response, JsonResponse]`. Seems that too much effort for that. 
However, it could be done when we will unify BadgerDoc responses

cakeinsauce: Yeah, it is not mandatory right now; JsonResponse inherit from Response and just setup media_type to json, that's it

cakeinsauce (approved):
Nicely done!

Issue #818, opened 2023-12-15
Restore unit-tests for all microservices


khyurri:
We need to restore unit-tests and add them into GitHub actions. 
Badgerdoc is going to use pytest as library to test, so all tests must be migrated to pytest. All tests must be added to GitHub actions.

- [x] annotation microservice (migrated to pytest, pre-commit and GitHub actions) [@cakeinsauce]
- [x] assets microservice (migrated to pytest, pre-commit and GitHub actions) [@cakeinsauce]
- [x] common libraries (migrated to pytest, pre-commit and GitHub actions) [@cakeinsauce]
- [ ] ~~convert microservice (migrated to pytest, pre-commit and GitHub actions)~~
- [x] jobs microservice (migrated to pytest, pre-commit and GitHub actions) [@cakeinsauce]
- [ ] ~~pipelines microservice (migrated to pytest, pre-commit and GitHub actions)~~
- [ ] ~~processing microservice (migrated to pytest, pre-commit and GitHub actions)~~
- [ ] ~~search microservice (migrated to pytest, pre-commit and GitHub actions)~~
- [x] taxonomy microservice (migrated to pytest, pre-commit and GitHub actions) [@cakeinsauce]
- [x] users microservice (migrated to pytest, pre-commit and GitHub actions) [@cakeinsauce]

PR #832, opened 2024-04-25
test(annotation): fix failing tests

cakeinsauce:
Fixed annotation service tests; removed/ignored kafka related tests due to its decommissioning
#818

### DIFF
--- a/.github/workflows/annotation.yml
+++ b/.github/workflows/annotation.yml
@@ -37,17 +37,6 @@ jobs:
     strategy:
       matrix:
         python-version: [ "3.8.15" ]
-    services:
-      postgres-postgresql:
-        image: postgres:13
-        ports:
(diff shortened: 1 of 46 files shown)


annotation/.env
```diff
@@ -17,9 +17,17 @@ S3_PROVIDER="minio"
 
 KEYCLOAK_HOST="http://dev2.badgerdoc.com/"
 
+ASSETS_SERVICE_SCHEME=http
```
khyurri, line 20: We have begun migrating to a single `.env` file for the docker-compose environment. It isn't logical to support both a global `.env` and individual ones for each microservice. If new variables are needed for unit testing, should they be mocked?

khyurri (approved):

khyurri (approved):

PR #835, opened 2024-05-01
ci: add build badgedoc github action

cakeinsauce:


.github/workflows/badgerdoc-build.yml
```diff
@@ -0,0 +1,16 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build services
+        run: make build_all
+      - name: Create .env config
+        run: cp ./.env.example ./.env
+      - name: Build the stack
+        run: docker-compose -f docker-compose-dev.yaml up -d
```
khyurri, line 16: I think for now `make build_all` is enough. Because we're not checking which services are up after `docker-compose -f docker-compose-dev.yaml up -d`, so it's just wasting CI time. What do you think?

khyurri (approved):

PR #839, opened 2024-05-03
feat: add Databricks support

khyurri:


.github/workflows/badgerdoc-build.yml
```diff
@@ -1,12 +1,12 @@
 name: badgerdoc build
 
 on:
-  push
+  pull_request
```
cakeinsauce, line 4: What do you want to achieve with this event? pull_request will only be triggered on pull_request modifications (not related to new commits).

Makefile
```diff
@@ -1,7 +1,11 @@
 _DOCKER_ ?= docker
 
+# Build all microservices
 build_all:  build_base build_base_3.12 build_annotation build_users build_convert build_jobs build_keycloak build_assets build_web  build_processing build_taxonomy clean
 
+# Build only BadgerDoc microservices
+build_badgerdoc: build_base build_base_3.12 build_annotation build_users build_convert build_jobs build_assets build_web  build_processing build_taxonomy clean
```
cakeinsauce, line 7: Do we use python base 3.12? Do we need it here at this moment?

jobs/alembic/versions/ef5c796d3c00_add_pipeline_engine.py
```diff
@@ -0,0 +1,28 @@
...
+revision = "ef5c796d3c00"
+down_revision = "1408c6c00120"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "job",
+        sa.Column("pipeline_engine", sa.String(length=255), nullable=True),
```
cakeinsauce, line 23: We create a new VARCHAR(255) column, but I did not notice any validation for the field length anywhere in the code. Are we good if user input is longer than 255, I think it will break

jobs/jobs/airflow_utils.py
```diff
@@ -0,0 +1,94 @@
...
+        password=AIRFLOW_PASSWORD,
+        host=AIRFLOW_URL,
+    )
+
+
+def get_api_instance():
+    configuration = get_configuration()
+
+    with client.ApiClient(configuration) as api_client:
+        yield api_client
```
cakeinsauce, line 38: I didn't notice any usage of this function anywhere in the code. Do we need it?

jobs/jobs/airflow_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    configuration = get_configuration()
+    with client.ApiClient(configuration) as api_client:
+        dag_api = DAGApi(api_client)
+        return dag_api.get_dags()
+
+
+def files_data_to_pipeline(
+    files_data: List[Dict[str, Any]],
+    job_id: int,
+) -> Iterable[pipeline.PipelineFile]:
```
cakeinsauce, line 51: What about `Generator[pipeline.PipelineFile, None, None` to be more precise?

jobs/jobs/airflow_utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def get_api_instance():
+    configuration = get_configuration()
+
+    with client.ApiClient(configuration) as api_client:
+        yield api_client
+
+
+async def get_dags():
```
cakeinsauce, line 41: Could you provide return type for this?

jobs/jobs/databricks_utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def get_client() -> WorkspaceClient:
+    url = f"{DATABRICKS_SERVICE_SCHEME}://{DATABRICKS_SERVICE_HOST}"
+    if DATABRICKS_SERVICE_PORT:
+        url = f"{url}:{DATABRICKS_SERVICE_PORT}"
+    return WorkspaceClient(host=url, token=os.environ["DATABRICKS_KEY"])
+
+
+def get_pipelines() -> Iterable[Pipeline]:
```
cakeinsauce, line 29: Maybe `Generator[Pipeline, None, None]` to be more precise?

jobs/jobs/databricks_utils.py
```diff
@@ -0,0 +1,83 @@
...
+            name=f"{pipeline_.settings.name}",
+        )
+
+
+async def run(
+    pipeline_id: str,
+    job_id: int,
+    files: List[pipeline.PipelineFile],
+    current_tenant: str,
+):
```
cakeinsauce, line 48:
```suggestion
) -> None:
```

jobs/jobs/databricks_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    logger.info(
+        "Running pipeline %s, job_id %s, current_tenant: %s with arguments %s",
+        pipeline_id,
+        files,
+        job_id,
+        current_tenant,
```
cakeinsauce, line 54: I think the order is incorrect:
```suggestion
    logger.info(
        "Running pipeline %s, job_id %s, current_tenant: %s with arguments %s",
        pipeline_id,
        job_id,
        current_tenant,
        files,
```

jobs/jobs/databricks_utils.py
```diff
@@ -0,0 +1,83 @@
...
+        return get_pipelines()
+
+    async def run(
+        self,
+        pipeline_id: str,
+        job_id: str,
+        files: List[pipeline.PipelineFile],
+        current_tenant: str,
+    ) -> None:
+        return await run(pipeline_id, job_id, files, current_tenant)
```
cakeinsauce, line 83:
```suggestion
        await run(pipeline_id, job_id, files, current_tenant)
```

jobs/jobs/main.py
```diff
@@ -49,17 +56,18 @@ async def create_job(
...
     if job_params.type == schemas.JobType.ExtractionJob:
-
         created_extraction_job = await create_job_funcs.create_extraction_job(
             extraction_job_input=job_params,  # type: ignore
             current_tenant=current_tenant,
             db=db,
             jw_token=jw_token,
         )
         if not job_params.is_draft:
+            logger.info("Running jobs")
```
cakeinsauce, line 70: Can we provide more info for this log or make it debug?

jobs/jobs/models.py
```diff
@@ -62,6 +62,7 @@ class ExtractionJob(Job):
     __abstract__ = True
     all_files_data = Column(JSONB)
     pipeline_id = Column(String(250))
+    pipeline_engine = Column(String(250))
```
cakeinsauce, line 65: We create a new VARCHAR(255) column in DB, but in code we have limit 250

jobs/jobs/pipeline.py
```diff
@@ -0,0 +1,47 @@
...
+
+@dataclass
+class AnyPipeline:
+    pass
+
+
+class BasePipeline(metaclass=ABCMeta):
+    @abstractmethod
+    async def list(self) -> List[AnyPipeline]:
+        raise NotImplementedError()
```
cakeinsauce, line 37: You can skip this as @abstractmethod takes care of it
```suggestion
        pass
```

jobs/jobs/run_job_funcs.py
```diff
@@ -19,27 +21,20 @@ async def run_extraction_job(
...
-
     db_service.update_job_mode(db, job_to_run, schemas.JobMode.Automatic)
     converted_files_data = utils.convert_files_data_for_inference(
         all_files_data=job_to_run.all_files_data,
         job_id=job_to_run.id,
         output_bucket=current_tenant,
     )
 
-    await utils.execute_pipeline(
+    logger.info("Starting external pipeline")
```
cakeinsauce, line 31: Can we provide more info in this log or make it debug?

jobs/jobs/schemas.py
```diff
@@ -343,3 +346,27 @@ class JobProgress(BaseModel):
...
+
+
+class Pipeline(BaseModel):
+    id: str
+    name: str
+    version: int = Field(default=0)
+    type: str = Field(default="airflow")
+    date: datetime
+    meta: Dict[str, Any]
+    steps: List[Any] = Field(default_factory=list)
```
cakeinsauce, line 368: BTW Pydantic handles default empty list correctly, without non-hashable object side effects.
```suggestion
    steps: List[Any] = []
```

jobs/jobs/schemas.py
```diff
@@ -343,3 +346,27 @@ class JobProgress(BaseModel):
...
+
+class PipelineEngineSupport(BaseModel):
+    data: List[PipelineEngine]
+
+
+class Pipeline(BaseModel):
+    id: str
+    name: str
+    version: int = Field(default=0)
+    type: str = Field(default="airflow")
```
cakeinsauce, line 365: You could do this
```suggestion
    version: int = 0
    type: str = "airflow"
```

jobs/jobs/utils.py
```diff
@@ -245,6 +248,50 @@ async def get_pipeline_instance_by_its_name(
     return response
 
 
+class UnsupportedEngine(Exception):
+    pass
+
+
+def files_data_to_pipeline_arg(
+    job_id: str, files_data: List[Dict[str, Any]]
+) -> Iterable[pipeline.PipelineFile]:
```
cakeinsauce, line 257: Maybe `Generator[pipeline.PipelineFile, None, None]` to be more precise?

jobs/jobs/airflow_utils.py
khyurri: We need to change to `Iterator` because of `__next__` support for comprehension

jobs/jobs/databricks_utils.py
khyurri: `run` must return pipeline id, but we're not specifying protocol right now, so I would skip this suggestion for now

Makefile
khyurri: You're right - we're not using it right now, however, base build already done and I don't know the reason why we shouldn't build it during CI

jobs/jobs/pipeline.py
khyurri: The `@abstractmethod` doesn't cover all scenarios of using an abstract class. For instance,
```
BasePipeline.list(AirflowPipeline())
```
with `pass` in the `list` method won't raise an exception.

While this example might seem odd, it's important to prevent improper usage of the created class hierarchy.

jobs/jobs/schemas.py
khyurri: For me `Explicit is better than implicit`

.github/workflows/badgerdoc-build.yml
khyurri: > What is your objective with this event?

My objective is to reduce the number of runs, as this process is quite CPU and network intensive 

> pull_request will only be triggered on pull_request modifications (not related to new commits).

This job will be executed when a commit is added to a pull_request.

jobs/jobs/run_job_funcs.py
khyurri: What would you suggest here to add? Feel free to add commit

jobs/jobs/main.py
```diff
@@ -18,6 +21,10 @@
 from jobs.config import KEYCLOAK_HOST, ROOT_PATH, API_current_version
 
 tenant = get_tenant_info(url=KEYCLOAK_HOST, algorithm="RS256", debug=True)
+logger = logging.getLogger(__name__)
+AIRFLOW_ENABLED = os.getenv("AIRFLOW_ENABLED", "").lower() == "true"
```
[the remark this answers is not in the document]
khyurri: I'm unsure if it makes sense at this point. What is the actual difference between an explicit comparison and calling `is_true`? Would it enhance code readability or boost the level of support? Does it even make sense to write a function for just one line? 

I would agree that a one-liner function can be used to improve readability, but I'm not sure whether it applies in this case.

jobs/jobs/run_job_funcs.py
cakeinsauce: `Starting external pipeline %s for job %s of %s tenant, %s engine`, job_to_run.pipeline_id, job_to_run.id, current_tenant, job_to_run.pipeline_engine

And Id make it debug tbh

jobs/jobs/airflow_utils.py
khyurri: Let's keep it until we finish airflow integration

PR #842, opened 2024-05-08
fix: return pipelines service back

cakeinsauce:


cakeinsauce:
Removed in 1.8.0, canceled

Issue #841, opened 2024-05-08
Add categories to job on annotation commit


khyurri:
BadgerDoc requires categories to be stored in the jobs table for display in the UI. Originally, BadgerDoc pipelines contained the available categories and assigned them to the job upon creation. However, the current BadgerDoc implementation excludes its own pipelines, necessitating a different method for committing categories. 

There are several possible approaches to commit categories to a job:
1. Explicitly during job creation - This approach is used for Annotation or Extraction + Annotation jobs, and we could employ it for Extraction jobs as well.
2. Explicitly through a BadgerDoc API call from an external pipeline.
3. Implicitly when the pipeline commits annotation results.

Options 1 and 2 are already available; however, if the annotation commit contains categories that have not been assigned, we encounter a bug where the categories are displayed as undefined. Also, explicit binding of categories to the job is a requirement for the Annotation and Annotation + Extraction jobs, because BadgerDoc needs to know which categories must be displayed during annotation. 

This task is to implement the implicit assignment of categories to a job when annotation results are committed.

### General Algorithm

When the annotation microservice commits, it must call the jobs microservice to assign the passed categories to the job. If the categories do not exist under the current tenant, the annotation microservice should not store the annotation results, and instead should send back an error. The jobs microservice should check the difference between the existing categories and those incoming, and then store the merged result. Since the categories in the jobs table are stored as JSONB, the calculation of the difference must be done in Python code. We need to lock the row for reading and modification until a new value is stored.

PR #843, opened 2024-05-08
feat: add categories to job on annotation commit

khyurri:
Closes: https://github.com/epam/badgerdoc/issues/841

### DIFF
--- a/annotation/annotation/annotations/resources.py
+++ b/annotation/annotation/annotations/resources.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Dict, List, Optional, Set
 from uuid import UUID
 
@@ -6,8 +7,14 @@ from sqlalchemy import and_, desc
 from sqlalchemy.orm import Session
 from tenant_dependency import TenantData
 
+from annotation.categories.services import (
+    combine_categories,
+    fetch_bunch_categories_db,
+)
 from annotation.database import get_db
 from annotation.errors import NoSuchRevisionsError
+from annotation.jobs.services import update_jobs_categories
+from annotation.microservice_communication import jobs_communication
 from annotation.microservice_communication.assets_communication import (
     get_file_path_and_bucket,
 )
@@ -34,7 +41,6 @@ from .main import (
     LATEST,
     accumulate_pages_info,
     add_search_annotation_producer,
-    check_if_kafka_message_is_needed,
     check_null_fields,
     check_task_pages,
     construct_annotated_doc,
@@ -52,6 +58,7 @@ router = APIRouter(
 )
 
 router.add_event_handler("startup", add_search_annotation_producer)
+logger = logging.getLogger(__name__)
 
 
 @router.post(
@@ -87,12 +94,12 @@ def post_annotation_by_user(
     fields with numbers of validated/failed_validation pages respectively.
     """
     if doc.user is None:
+        detail = (
+            "Field user should not be null, when saving annotation by user."
+        )
         raise HTTPException(
             status_code=400,
-            detail=(
-                "Field user should not be null, "
-                "when saving annotation by user."
-            ),
+            detail=detail,
         )
     check_null_fields(doc)
 
@@ -215,15 +222,6 @@ def post_annotation_by_user(
             detail=f"Cannot assign similar documents: {err}",
         ) from err
     db.commit()
-    if False:  # Kafka needs to be removed
-        check_if_kafka_message_is_needed(
-            db,
-            latest_doc,
-            new_annotated_doc,
-            task.job_id,
-            task.file_id,
-            x_current_tenant,
-        )
     return AnnotatedDocSchema.from_orm(new_annotated_doc)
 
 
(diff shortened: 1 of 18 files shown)


annotation/annotation/annotations/resources.py
```diff
@@ -90,8 +92,7 @@ def post_annotation_by_user(
         raise HTTPException(
             status_code=400,
             detail=(
-                "Field user should not be null, "
-                "when saving annotation by user."
+                "Field user should not be null, " "when saving annotation by user."
```
cakeinsauce, line 95:
```suggestion
                "Field user should not be null, when saving annotation by user."
```

annotation/annotation/annotations/resources.py
```diff
@@ -288,6 +278,16 @@ def post_annotation_by_pipeline(
...
     )
+    # Add categories to jobs microservice for any commit to keep
+    # it consistent (https://github.com/epam/badgerdoc/issues/841)
+    categories = combine_categories(doc.categories, doc.pages)
+    category_objects = fetch_bunch_categories_db(
+        db, categories, x_current_tenant, root_parents=True
+    )
+    update_jobs_categories(db, job_id, category_objects)
+    jobs_communication.append_job_categories(
+        job_id, list(categories), x_current_tenant, token.token
```
cakeinsauce, line 289: `combine_categories` already returns list of categories, I don't think we need explicit type cast here
```suggestion
        job_id, categories, x_current_tenant, token.token
```

annotation/annotation/annotations/resources.py
```diff
@@ -539,10 +526,7 @@ def get_annotation_for_given_revision(
     responses={
         500: {"model": ConnectionErrorSchema},
     },
-    summary=(
-        "Get all users revisions (or pipeline revision) "
-        "for particular pages."
-    ),
+    summary=("Get all users revisions (or pipeline revision) " "for particular pages."),
```
cakeinsauce, line 529:
```suggestion
    summary=("Get all users revisions (or pipeline revision) for particular pages."),
```

annotation/annotation/annotations/resources.py
```diff
@@ -553,9 +537,7 @@ def get_all_revisions(
     user_id: Optional[UUID] = Query(
         None,
         example="1843c251-564b-4c2f-8d42-c61fdac369a1",
-        description=(
-            "Required in case job validation " "type is extensive_coverage"
-        ),
+        description=("Required in case job validation " "type is extensive_coverage"),
```
cakeinsauce, line 540:
```suggestion
        description=("Required in case job validation type is extensive_coverage"),
```

jobs/jobs/categories.py
```diff
@@ -0,0 +1,55 @@
...
+logger = logging.getLogger(__name__)
+
+
+async def prepare_for_update(
+    current_job: models.CombinedJob,
+    new_job: schemas.JobParamsToChange,
+    current_tenant: str,
+    jw_token: str,
+) -> Tuple[List[str], List[schemas.CategoryLinkInput]]:
+    logger.info("Prepare categories for update")
```
cakeinsauce, line 17: Do we need this log? We log later certain categories for certain actions

jobs/jobs/db_service.py
```diff
@@ -145,10 +145,13 @@ def get_all_jobs(db: Session) -> List[Dict[str, Any]]:
...
 def get_job_in_db_by_id(
-    db: Session, job_id: int
+    db: Session, job_id: int, with_lock=False
 ) -> Union[dbm.CombinedJob, Any]:
     """Getting hold on a job in the database by its id"""
-    job_needed = db.query(dbm.CombinedJob).get(job_id)
+    if with_lock:
+        job_needed = db.query(dbm.CombinedJob).get(job_id)
+    else:
+        job_needed = db.query(dbm.CombinedJob).with_for_update().get(job_id)
```
cakeinsauce, line 154: Shouldn't it be vice versa?

annotation/annotation/annotations/resources.py
khyurri: `combine_categories` function returns a `Set[str]`. However, we need to serialize it for passing as a JSON string, so here we're explicitly converting from set to list. It seems that your suggestion might break the downstream code.

jobs/jobs/categories.py
khyurri: This log message could be removed. However, if something inside the function raises an exception, the last message in the log stream, 'Prepare categories for update,' can easily help us identify the last called function. 

I agree that we could find the complete list of calls in the stack trace, but this message serves as an additional logging point.

jobs/jobs/db_service.py
khyurri: Great catch!

jobs/jobs/db_service.py
```diff
@@ -145,10 +145,13 @@ def get_all_jobs(db: Session) -> List[Dict[str, Any]]:
 
 
 def get_job_in_db_by_id(
-    db: Session, job_id: int
+    db: Session, job_id: int, with_lock=False
```
cakeinsauce, line 148:
```suggestion
    db: Session, job_id: int, with_lock: bool = False
```

PR #845, opened 2024-05-10
feat: added pre signed s3 urls in job service

### LINKED ISSUE #840: Add an S3 Signed URL as an Argument to Pipelines
Currently, BadgerDoc executes pipelines with the `files_data` parameter under the assumption that the pipeline engine can access the S3 file without any extra configuration (stores credentials on its own side). However, we need to enable pipeline engines to download S3 files without storing authentication credentials on their side. Therefore, this task involves adding the capability to generate S3 signed URLs as passed arguments to the pipeline manager

The PR https://github.com/epam/badgerdoc/pull/839 adds functionality to execute pipelines with additional arguments using the dataclass: 

```
@dataclass
class PipelineFile:
    bucket: str
    input: PipelineFileInput
    input_path: str
    pages: List[int]
    output_path: Optional[str] = None
    s3_signed_url: Optional[str] = None
    annotation_id: Optional[str] = None
```

The `s3_signed_url` needs to be filled with the generated Signed Url if BadgerDoc is configured with the parameter: `JOBS_RUN_PIPELINES_WITH_SIGNED_URL=True`. This value can only be set to True if `S3_PROVIDER` is configured as `aws_iam`. By default, `JOBS_RUN_PIPELINES_WITH_SIGNED_URL` is set to False.

### What needs to be changed

1. Add `JOBS_RUN_PIPELINES_WITH_SIGNED_URL=` to `.env.example`

### What needs to be changed

1. Add `JOBS_RUN_PIPELINES_WITH_SIGNED_URL=` to the `.env.example`. 

### What needs to be additionally checked

Given that we're expecting a huge (almost unlimited) number of documents to be passed as S3 signed URLs, `aioboto3` library integration into the `jobs` microservice could be considered to speed up the process of generating signed URLs.

SunnyCapt:
close #840


jobs/jobs/s3.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def create_pre_signed_s3_url(
+    bucket: str, path: str,
+    method: Literal["GET"] = "GET",
+    expires: timedelta = timedelta(days=config.S3_PRE_SIGNED_EXPIRES_DAYS),
+    client: Minio = None,  # type: ignore
+) -> str:
+    if client is None:
+        from .main import minio_client as client
```
khyurri, line 63: Could you please explain why you've postponed client loading? I mean, it might not be expected that a function loads something from a module. If it's possible to avoid it, it's generally better not to do it.

jobs/.env
```diff
@@ -10,3 +10,11 @@ POSTGRES_PORT=5432
 POSTGRES_USER=postgres
 POSTGRES_PASSWORD=postgres
 POSTGRES_DB=badgerdoc
+
+# S3 settings
```
khyurri, line 14: Let's avoid adding env variables into local `.env` files, better to use common `.env.example` from root. We're planning to remove all `.env`

.env.example
```diff
@@ -36,6 +36,10 @@ MINIO_DEFAULT_BUCKETS=local
 
 S3_PREFIX=
 
+# pre signed urls expire in 48 hours
+S3_PRE_SIGNED_EXPIRES_HOURS=48
+JOBS_RUN_PIPELINES_WITH_SIGNED_URL=False
```
khyurri, line 41: Let's move it to the `JOBS_` section

jobs/jobs/utils.py
```diff
@@ -266,6 +269,20 @@ def files_data_to_pipeline_arg(
...
+    async def fill(file):
+        file.s3_signed_url = await create_pre_signed_s3_url(
+            bucket=file.bucket, path=file.input_path
+        )
+
+    if not JOBS_RUN_PIPELINES_WITH_SIGNED_URL:
+        return files
+
+    tasks = [fill(f) for f in files]
+    await asyncio.gather(*tasks)
```
khyurri, line 282: I think we don't need it to be run in parallel here, it's ok to run 1 by 1, but `await` results to allow other `coroutines` do their job

SunnyCapt: could you please explain why you think so?

khyurri (approved):

PR #846, opened 2024-05-11
feat: black & isort common workflow

cakeinsauce:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v44
+      - name: List changed files
+        run: |
+          echo Changed files: ${{ steps.changed-files.outputs.all_changed_files }}
+      - name: Install python dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install black==24.4.2 pylint==3.1.0 mypy==1.10.0 isort==5.13.2
```
khyurri, line 31: I would not recommend installing linters in CI because we need to have the same version installed locally. 
Should we specify development dependencies in the `pyproject.toml` file for Poetry?

cakeinsauce: We can, but we cannot install deps from pyproject.toml using pip. Im not sure if we should use poetry here...

khyurri: Why not? Poetry is our requirement for every microservice

khyurri:
Let's keep in this PR only black and isort. Then we will be able to run whole project using black and isort to exclude format changes in next PR's

khyurri (approved):
Thank you!

PR #865, opened 2024-06-06
fix: replace boto3 with aioboto3

cakeinsauce:


khyurri:
It seems that we can close this PR because most of the migration work into badgerdoc_storage is in the branch https://github.com/epam/badgerdoc/tree/feature/add_blob_storage

cakeinsauce:
> It seems that we can close this PR because most of the migration work into badgerdoc_storage is in the branch https://github.com/epam/badgerdoc/tree/feature/add_blob_storage

Good point, closing..

PR #889, opened 2024-08-07
test: add test for validate_task_info function in tasks module

BogicevicMilan:
Added a unit test for the validate_task_info function located in the tasks module of annotations.

### DIFF
--- a/annotation/tests/test_tasks_services.py
+++ b/annotation/tests/test_tasks_services.py
@@ -1,0 +1,52 @@
+from unittest.mock import Mock, patch
+
+import pytest
+from sqlalchemy.orm import Session
+
+from annotation.errors import FieldConstraintError
+from annotation.jobs.services import ValidationSchema
+from annotation.tasks.services import validate_task_info
+
+
+@pytest.fixture
+def db_session():
+    """Fixture for creating a mock database session."""
+    return Mock(spec=Session)
+
+
+@pytest.fixture
+def task_info():
+    """Fixture for creating a mock task_info dictionary."""
+    return {"is_validation": True}
+
+
+@pytest.mark.parametrize(
+    "validation_type, is_validation, should_raise_error",
+    [
+        (ValidationSchema.validation_only, False, True),
+        (ValidationSchema.validation_only, True, False),
+        ("other_type", False, False),
+        ("other_type", True, False),
+    ],
+)
+def test_validate_task_info(
+    db_session, task_info, validation_type, is_validation, should_raise_error
+):
+    task_info["is_validation"] = is_validation
+
+    if should_raise_error:
+        with pytest.raises(FieldConstraintError):
+            validate_task_info(db_session, task_info, validation_type)
+    else:
+        with patch(
+            "annotation.tasks.services.validate_users_info"
+        ) as mock_validate_users_info, patch(
+            "annotation.tasks.services.validate_files_info"
+        ) as mock_validate_files_info:
+            validate_task_info(db_session, task_info, validation_type)
+            mock_validate_users_info.assert_called_once_with(
+                db_session, task_info, validation_type
+            )
+            mock_validate_files_info.assert_called_once_with(
+                db_session, task_info
+            )
--- a/annotation/tests/test_tasks_services.py
+++ b/annotation/tests/test_tasks_services.py
@@ -1,52 +1,158 @@
-from unittest.mock import Mock, patch
+from unittest.mock import Mock, call, patch
 
+from annotation.models import File
 import pytest
 from sqlalchemy.orm import Session
 
 from annotation.errors import FieldConstraintError
 from annotation.jobs.services import ValidationSchema
-from annotation.tasks.services import validate_task_info
+from annotation.tasks.services import validate_files_info, validate_task_info, validate_users_info
 
 
-@pytest.fixture
+TEST_PARAMS_INVALID_FILES = [
+    (
+        {
+            "file_id": 1,
+            "job_id": 2,
+            "pages": [1, 2, 4],
+        },
+        Mock(spec=File, pages_number=3),
+        "pages ({4}) do not belong to file"
+    ),
+    (
+        {
+            "file_id": 1,
+            "job_id": 2,
+            "pages": [1, 2, 3],
+        },
+        None,
+        "file with id 1 is not assigned for job 2"
+    ),
+]
+
+
+@pytest.fixture(scope="function")
 def db_session():
-    """Fixture for creating a mock database session."""
     return Mock(spec=Session)
 
-
-@pytest.fixture
+@pytest.fixture(scope="function")
 def task_info():
-    """Fixture for creating a mock task_info dictionary."""
-    return {"is_validation": True}
+    return {
+        "is_validation": True,
+        "user_id": 1,
+        "job_id": 2,
+    }
 
 
 @pytest.mark.parametrize(
-    "validation_type, is_validation, should_raise_error",
-    [
-        (ValidationSchema.validation_only, False, True),
-        (ValidationSchema.validation_only, True, False),
-        ("other_type", False, False),
-        ("other_type", True, False),
-    ],
+    ("validation_type", "is_validation"),
+    (
+        (ValidationSchema.validation_only, True),
+        ("other_type", False),
+        ("other_type", True),
+    ),
 )
-def test_validate_task_info(
-    db_session, task_info, validation_type, is_validation, should_raise_error
+def test_validate_task_info_positive_case(
+    db_session: Session, task_info: dict, validation_type: ValidationSchema, is_validation: bool
 ):
     task_info["is_validation"] = is_validation
 
-    if should_raise_error:
-        with pytest.raises(FieldConstraintError):
-            validate_task_info(db_session, task_info, validation_type)
+    with patch(
+        "annotation.tasks.services.validate_users_info"
+    ) as mock_validate_users_info, patch(
+        "annotation.tasks.services.validate_files_info"
+    ) as mock_validate_files_info:
+        validate_task_info(db_session, task_info, validation_type)
+        mock_validate_users_info.assert_called_once_with(
+            db_session, task_info, validation_type
+        )
+        mock_validate_files_info.assert_called_once_with(
+            db_session, task_info
+        )
+
+
+@pytest.mark.parametrize(
+    ("validation_type", "is_validation"),
+    (
+        (ValidationSchema.validation_only, False),
+    ),
+)
+def test_validate_task_info_negative_case(
+    db_session: Session,task_info: dict, validation_type: ValidationSchema, is_validation: bool
+):
+    task_info["is_validation"] = is_validation
+
+    with pytest.raises(FieldConstraintError):
+        validate_task_info(db_session, task_info, validation_type)
+
+
+@pytest.mark.parametrize(
+    "validation_type, is_validation, mock_function, mock_return_value",
+    (
+        (ValidationSchema.cross, True, "check_cross_annotating_pages", True),
+        (ValidationSchema.validation_only, True, None, True),
+        (ValidationSchema.validation_only, False, None, True),
+    )
+)
+def test_validate_users_info_positive_case(
+    db_session: Session, task_info: dict, validation_type: ValidationSchema, is_validation: bool, mock_function: str, mock_return_value: bool
+):
+    task_info["is_validation"] = is_validation
+    
+    if mock_function:
+        with patch(f"annotation.tasks.services.{mock_function}") as mock_func:
+            validate_users_info(db_session, task_info, validation_type)
+            mock_func.assert_called_once_with(db_session, task_info)
     else:
-        with patch(
-            "annotation.tasks.services.validate_users_info"
-        ) as mock_validate_users_info, patch(
-            "annotation.tasks.services.validate_files_info"
-        ) as mock_validate_files_info:
-            validate_task_info(db_session, task_info, validation_type)
-            mock_validate_users_info.assert_called_once_with(
-                db_session, task_info, validation_type
-            )
-            mock_validate_files_info.assert_called_once_with(
-                db_session, task_info
-            )
+        db_session.query().filter_by().first.return_value = mock_return_value
+        validate_users_info(db_session, task_info, validation_type)
+
+
+@pytest.mark.parametrize(
+    ("is_validation", "expected_error_message"),
+    (
+        (True, "user 1 is not assigned as validator for job 2"),
+        (False,"user 1 is not assigned as annotator for job 2")
+    )
+)
+def test_validate_users_info_negative_case(db_session: Session, task_info: dict, is_validation: bool, expected_error_message: str):
+    task_info["is_validation"] = is_validation
+    db_session.query().filter_by().first.return_value = None
+    with pytest.raises(FieldConstraintError) as excinfo:
+        validate_users_info(db_session, task_info, ValidationSchema.validation_only)
+    assert expected_error_message in str(excinfo.value)
+
+
+def test_validate_files_info_positive_case(db_session: Session):
+    task_info = {
+        "file_id": 1,
+        "job_id": 2,
+        "pages": [1, 2, 3],
+    }
+    
+    mock_file = Mock(spec=File)
+    mock_file.pages_number = 3
+    
+    mock_query = db_session.query.return_value
+    mock_query.filter_by.return_value.first.return_value = mock_file
+        
+    validate_files_info(db_session, task_info)
+    
+    assert db_session.query.call_count == 1
+    db_session.query.assert_has_calls([
+        call(File),
+    ])
+    mock_query.filter_by.assert_called_once_with(file_id=1, job_id=2)
+    
+
+@pytest.mark.parametrize(
+    ("task_info", "mock_file", "expected_error_message"),
+    TEST_PARAMS_INVALID_FILES
+)
+def test_validate_files_info_negative_case(db_session: Session, task_info: dict, mock_file: Mock, expected_error_message: str):
+    db_session.query().filter_by().first.return_value = mock_file
+
+    with pytest.raises(FieldConstraintError) as excinfo:
+        validate_files_info(db_session, task_info)
+    
+    assert expected_error_message in str(excinfo.value)
--- a/annotation/.pre-commit-config.yaml
+++ b/annotation/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 fail_fast: true
 repos:
   - repo: https://github.com/pycqa/isort
-    rev: 5.9.2
+    rev: 5.12.0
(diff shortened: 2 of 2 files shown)


annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+
+from annotation.errors import FieldConstraintError
+from annotation.jobs.services import ValidationSchema
+from annotation.tasks.services import validate_task_info
+
+
+@pytest.fixture
+def db_session():
+    """Fixture for creating a mock database session."""
+    return Mock(spec=Session)
```
djo753, line 14: Good use of Mock object. We should check if there exists a similar fixture inside of `conftest.py`, so that we don't duplicate code.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+
+    if should_raise_error:
+        with pytest.raises(FieldConstraintError):
+            validate_task_info(db_session, task_info, validation_type)
+    else:
+        with patch(
+            "annotation.tasks.services.validate_users_info"
+        ) as mock_validate_users_info, patch(
+            "annotation.tasks.services.validate_files_info"
+        ) as mock_validate_files_info:
```
djo753, line 45: Great idea with patch()

djo753 (commented):
The tests run and pass. We should check the existence `db_session()` fixture somewhere else in our codebase.

annotation/tests/test_tasks_services.py
BogicevicMilan: There isn't something like this in conftest.py, so I created it as a @pytest.fixture to avoid code duplication because I will need the DB session in many tests. After writing the upcoming tests, I might extend this function further.

annotation/tests/test_tasks_services.py
BogicevicMilan: Thanks.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+
+@pytest.fixture
+def db_session():
+    """Fixture for creating a mock database session."""
+    return Mock(spec=Session)
+
+
+@pytest.fixture
+def task_info():
+    """Fixture for creating a mock task_info dictionary."""
```
cakeinsauce, line 19: No need for such docstrings

annotation/tests/test_tasks_services.py
cakeinsauce, line 14: check out teams chat regarding fixtures

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+    return {"is_validation": True}
```
cakeinsauce, line 20: check out teams chat regarding fixtures

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@pytest.mark.parametrize(
+    "validation_type, is_validation, should_raise_error",
```
cakeinsauce, line 24: put them in tuple, please. more readable

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+    [
+        (ValidationSchema.validation_only, False, True),
+        (ValidationSchema.validation_only, True, False),
+        ("other_type", False, False),
+        ("other_type", True, False),
+    ],
```
cakeinsauce, line 30: whenever you see that collection wont change during its lifetime - use tuples.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+)
+def test_validate_task_info(
+    db_session, task_info, validation_type, is_validation, should_raise_error
```
cakeinsauce, line 33: add type hints, please

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,52 @@
...
+):
+    task_info["is_validation"] = is_validation
+
+    if should_raise_error:
```
cakeinsauce, line 37: it is considered a bad practice add logic into test. Feel free to separate test into different. Especially for good and bad cases

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+from annotation.models import File
+import pytest
+from sqlalchemy.orm import Session
+
+from annotation.errors import FieldConstraintError
+from annotation.jobs.services import ValidationSchema
+from annotation.tasks.services import validate_files_info, validate_task_info, validate_users_info
+
+
+TEST_PARAMS_INVALID_FILES = [
```
cakeinsauce, line 12: Move this constant into appropriate test, no need for it here

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
+from unittest.mock import Mock, call, patch
+
+from annotation.models import File
+import pytest
+from sqlalchemy.orm import Session
+
+from annotation.errors import FieldConstraintError
+from annotation.jobs.services import ValidationSchema
+from annotation.tasks.services import validate_files_info, validate_task_info, validate_users_info
```
cakeinsauce, line 9: please fix your sorts using pre-commit or just isort

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+            "job_id": 2,
+            "pages": [1, 2, 3],
+        },
+        None,
+        "file with id 1 is not assigned for job 2"
+    ),
+]
+
+
+@pytest.fixture(scope="function")
```
cakeinsauce, line 34: Default scope for fixture is function, no need to write it here
```suggestion
@pytest.fixture
```

annotation/tests/test_tasks_services.py
cakeinsauce: use patch for this. If you want just a mock object, then no need for this fixture, move it into test

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+@pytest.mark.parametrize(
+    ("validation_type", "is_validation"),
+    (
+        (ValidationSchema.validation_only, True),
+        ("other_type", False),
+        ("other_type", True),
+    ),
+)
+def test_validate_task_info_positive_case(
+    db_session: Session, task_info: dict, validation_type: ValidationSchema, is_validation: bool
```
cakeinsauce, line 56: db_session is Mock type

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+            db_session, task_info
+        )
+
+
+@pytest.mark.parametrize(
+    ("validation_type", "is_validation"),
+    (
+        (ValidationSchema.validation_only, False),
+    ),
+)
```
cakeinsauce, line 79: No need for a single parametrize, you can remove it

annotation/tests/test_tasks_services.py
cakeinsauce: and also use yield instead of return

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+]
+
+
+@pytest.fixture(scope="function")
+def db_session():
+    return Mock(spec=Session)
+
+@pytest.fixture(scope="function")
+def task_info():
+    return {
```
cakeinsauce, line 40: yield

annotation/tests/test_tasks_services.py
cakeinsauce: Always specify nested types in collections, not just dict

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+@pytest.mark.parametrize(
+    ("validation_type", "is_validation"),
+    (
+        (ValidationSchema.validation_only, False),
+    ),
+)
+def test_validate_task_info_negative_case(
+    db_session: Session,task_info: dict, validation_type: ValidationSchema, is_validation: bool
```
cakeinsauce, line 81:
```suggestion
    db_session: Session, task_info: dict, validation_type: ValidationSchema, is_validation: bool
```

cakeinsauce: and specify dict nested types

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+):
+    task_info["is_validation"] = is_validation
+
+    with pytest.raises(FieldConstraintError):
+        validate_task_info(db_session, task_info, validation_type)
+
+
+@pytest.mark.parametrize(
+    "validation_type, is_validation, mock_function, mock_return_value",
```
cakeinsauce, line 90: str -> tuple

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+    (
+        (ValidationSchema.cross, True, "check_cross_annotating_pages", True),
+        (ValidationSchema.validation_only, True, None, True),
+        (ValidationSchema.validation_only, False, None, True),
+    )
+)
+def test_validate_users_info_positive_case(
+    db_session: Session, task_info: dict, validation_type: ValidationSchema, is_validation: bool, mock_function: str, mock_return_value: bool
```
cakeinsauce, line 98: session -> mock, dict[<type>, <type>]

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+    with pytest.raises(FieldConstraintError):
+        validate_task_info(db_session, task_info, validation_type)
+
+
+@pytest.mark.parametrize(
+    "validation_type, is_validation, mock_function, mock_return_value",
+    (
+        (ValidationSchema.cross, True, "check_cross_annotating_pages", True),
+        (ValidationSchema.validation_only, True, None, True),
+        (ValidationSchema.validation_only, False, None, True),
```
cakeinsauce, line 94: Your last value `mock_return_value` does not change in parametrize, so you can omit it

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+):
+    task_info["is_validation"] = is_validation
+    
+    if mock_function:
+        with patch(f"annotation.tasks.services.{mock_function}") as mock_func:
+            validate_users_info(db_session, task_info, validation_type)
+            mock_func.assert_called_once_with(db_session, task_info)
+    else:
+        db_session.query().filter_by().first.return_value = mock_return_value
+        validate_users_info(db_session, task_info, validation_type)
```
cakeinsauce, line 108: Remove if logic from a test, please. Either adjust test in parametrize, or separate it into two tests. Whenever you have logic inside a test, test becomes a possible fail point

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+@pytest.mark.parametrize(
+    ("is_validation", "expected_error_message"),
+    (
+        (True, "user 1 is not assigned as validator for job 2"),
+        (False,"user 1 is not assigned as annotator for job 2")
```
cakeinsauce, line 115: specify `validator` and `annotator` in parametrize instead of the full message. you can format the full message later

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+    )
+)
+def test_validate_users_info_negative_case(db_session: Session, task_info: dict, is_validation: bool, expected_error_message: str):
```
cakeinsauce, line 118: session -> mock, dict -> more specific

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+    task_info["is_validation"] = is_validation
+    db_session.query().filter_by().first.return_value = None
+    with pytest.raises(FieldConstraintError) as excinfo:
+        validate_users_info(db_session, task_info, ValidationSchema.validation_only)
+    assert expected_error_message in str(excinfo.value)
```
cakeinsauce, line 123: https://docs.pytest.org/en/stable/how-to/assert.html#matching-exception-messages

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def test_validate_files_info_positive_case(db_session: Session):
```
cakeinsauce, line 126: session -> mock

cakeinsauce: no need to say `positive_case`. If you dont say it is edge/error case, we assume it is positive case

annotation/tests/test_tasks_services.py
cakeinsauce: Could you specify more explanatory name instead of `negative_case`? Like what you do here, when such cases appear

annotation/tests/test_tasks_services.py
cakeinsauce: remove `positive_case`

annotation/tests/test_tasks_services.py
cakeinsauce: Instead of `negative_case` specify explanatory name of what you do here/when such cases appear

annotation/tests/test_tasks_services.py
cakeinsauce: remove `positive_case`

annotation/tests/test_tasks_services.py
BogicevicMilan: Should I replace positive_case and negative_case with good_case and bad_case throughout the code?

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+    db_session.query.assert_has_calls([
+        call(File),
+    ])
+    mock_query.filter_by.assert_called_once_with(file_id=1, job_id=2)
+    
+
+@pytest.mark.parametrize(
+    ("task_info", "mock_file", "expected_error_message"),
+    TEST_PARAMS_INVALID_FILES
+)
```
cakeinsauce, line 151: Fix after previous comment is resolved (related to TEST_PARAMS_INVALID_FILES)

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+def test_validate_files_info_negative_case(db_session: Session, task_info: dict, mock_file: Mock, expected_error_message: str):
```
cakeinsauce, line 152: negative_case -> explanatory name

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+    db_session.query().filter_by().first.return_value = mock_file
+
+    with pytest.raises(FieldConstraintError) as excinfo:
+        validate_files_info(db_session, task_info)
+    
+    assert expected_error_message in str(excinfo.value)
```
cakeinsauce, line 158: https://docs.pytest.org/en/stable/how-to/assert.html#matching-exception-messages

cakeinsauce: I think your pre-commit doesnt work, it wouldn't let you commit with such formattings

BogicevicMilan: I used LF instead of CRLF.

annotation/tests/test_tasks_services.py
BogicevicMilan: Fixed :)

annotation/tests/test_tasks_services.py
BogicevicMilan: Fixed :)

annotation/tests/test_tasks_services.py
BogicevicMilan: Should I use this as a parameterized test without the constant, or should the constant be placed directly before the function that uses it? I went with the latter approach.

annotation/tests/test_tasks_services.py
BogicevicMilan: Fixed :)

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,158 @@
...
+        "file with id 1 is not assigned for job 2"
+    ),
+]
+
+
+@pytest.fixture(scope="function")
+def db_session():
+    return Mock(spec=Session)
+
+@pytest.fixture(scope="function")
```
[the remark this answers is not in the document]
BogicevicMilan: Fixed :)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok :)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok ;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok :)

annotation/tests/test_tasks_services.py
BogicevicMilan: Fixed ;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok ;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok :)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: ?

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok ;)

BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Fixed;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Fixed ;)

PR #893, opened 2024-08-08
test(annotation): cover distribution/main.py

UkiStefanovic:
This Pull request adds unit tests for multiple functions from the distribution/main.py module. Functions covered with this pull request are:
-  choose_validators_users, 
- prepare_users, 
- find_users_share_loads, 
- find_equal_files, 
- find_small_files, 
- prepare_response.

You can find the unit tests for these function in /tests/test_distribution.py


annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
     expected: List[int],
 ) -> None:
     assert get_page_number_combinations(file_pages, user_pages) == expected
+
+
+# ADDITIONAL TESTS
+
```
Negoje, line 1289: We should check if similar mock objects already exist, so that we do not create unnecessary new objects.

Negoje (commented):
All tests are green. Good job!

annotation/.pre-commit-config.yaml
```diff
@@ -1,7 +1,10 @@
+# https://pre-commit.com
+# https://pre-commit.com/hooks.html
+
```
cakeinsauce, line 3: No need for that

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
     expected: List[int],
 ) -> None:
     assert get_page_number_combinations(file_pages, user_pages) == expected
+
+
+# ADDITIONAL TESTS
```
cakeinsauce, line 1288: No need for this comment

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+    {
+        "user_id": "405ef0e2-b53e-4c18-bf08-c0871615d99d",
+        "default_load": 100,
+        "overall_load": 0,
+    },
+    {
+        "user_id": "ba3e0ccf-8661-4ed3-892b-7c291160f631",
+        "default_load": 96,
+        "overall_load": 0,
+    },
```
cakeinsauce, line 1300: Do you need so much users for your tests? Usually you would have something normal and edge cases, apply this principle here as well. And make it a fixture, please. Same applies to ANNOTATION_TASKS_CHOOSE_VALIDATORS

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+            "user_id": "c080408e-0077-4c20-b520-bd4b1541ca59",
+            "default_load": 0,
+            "overall_load": 50,
+        },
+    ],
+    [
+        {
+            "user_id": "6e2ce9ac-2fe8-4cc7-bdcd-bac90faa9247",
+            "default_load": 95,
+            "overall_load": 0,
```
cakeinsauce, line 1390: Expected output is always related to a certain test, no need to keep in in a module level. Additionally, all data you prepared above should be used in test directly if you dont utilize in multiple places

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+    "annotators,"
+    "validators,"
+    "annotation_tasks,"
+    "expected_output",
+    [
+        (
+            False,
+            ValidationSchema.cross,
+            USERS[0:2] + USERS[7:],
+            USERS[2:7],
```
cakeinsauce, line 1428: This is not really readable. After you change testing data according to my comments above, change this as well. And do you really need so much users to test the functionally? Take a bare minimum to cover different cases

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+            True,
+            ValidationSchema.validation_only,
+            USERS[0:2] + USERS[7:],
+            USERS[2:7],
+            ANNOTATION_TASKS_CHOOSE_VALIDATORS,
+            EXPECTED_OUTPUT[1],
+        ),
+    ],
+)
+@pytest.mark.unittest
```
cakeinsauce, line 1450: No need for this

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+def test_choose_validators_users(
+    split_multipage_doc_setup,
+    validation_type,
+    annotators,
+    validators,
+    annotation_tasks,
+    expected_output,
```
cakeinsauce, line 1457: Add type hints, please

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+):
+    # SETUP
+    # Set environment variable SPLIT_MULTIPAGE_DOC
```
cakeinsauce, line 1460: No need for such comments, code is pretty explanatory. Whenever you think about writing a new comment, ask yourself - is my code understandable without comments? If not, most likely you should change a code

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+    if split_multipage_doc_setup:
+        os.environ["SPLIT_MULTIPAGE_DOC"] = "1"
+    else:
+        del os.environ["SPLIT_MULTIPAGE_DOC"]
```
cakeinsauce, line 1464: You should never mock env vars like that (any mutable object). It brings a lot of side effects, you don't want to mess with that later.
Check out this
https://docs.python.org/3/library/unittest.mock.html#patch-dict

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+
+    output = choose_validators_users(
+        validation_type, annotators, validators, annotation_tasks
+    )
+
+    # Checking is lists have the same elements
+    # expected_output is already sorted by 'user_id'
```
cakeinsauce, line 1471: No need for this comment

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1283,191 @@ def test_get_page_number_combinations(
...
+    sorted_output = sorted(output, key=lambda x: x["user_id"])
```
cakeinsauce, line 1472: Could you elaborate why we sort output? Interfere with output for testing is not the best practice. Add logic to a test overall

UkiStefanovic: Yes, I've noticed that the function returns the list of same elements, but it shuffles them every time. So my idea was to sort the output list because then sorted_output would always be the same list for every test run, but output wouldn't. Ideally, we would like the function to return a set, and then I wouldn't have sorted the output, I would have just done assert output==expected, but because output list is mixed every run, I sorted it. I'm currently changing the test cases to be very simple and possibly will not have sorting in the teardown. Should I never sort the output if I just need to check if elements of two lists are the same? And if so, why?

UkiStefanovic: Nevermind, this code is now irrelevant. Check out my fixed test

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1284,48 @@ def test_get_page_number_combinations(
...
+        ),
+        (
+            "1",
+            ValidationSchema.hierarchical,
+            [{"user_id": "0"}, {"user_id": "1"}],
+            [{"user_id": "2"}],
+            [{"user_id": "1"}],
+            [{"user_id": "2"}],
+        ),
+    ],
```
cakeinsauce, line 1315: list -> tuple (outer one)

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1284,48 @@ def test_get_page_number_combinations(
...
+def test_choose_validators_users(
+    split_multipage_doc_setup: str,
+    validation_type: ValidationSchema,
+    annotators: List[User],
+    validators: List[User],
+    annotation_tasks: List[Task],
+    expected_output: List[User],
+):
+    with patch.dict(
+        "os.environ", {"SPLIT_MULTIPAGE_DOC": split_multipage_doc_setup}
```
cakeinsauce, line 1326: Patching os.environ is not the best practice. Patching env vars overall. Let's patch annotation.distribution.main.SPLIT_MULTIPAGE_DOC value itself

cakeinsauce (approved):
LGTM!

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1286,97 @@ def test_get_page_number_combinations(
...
+        ),
+    ),
+)
+def test_choose_validators_users(
+    split_multipage_doc_setup: str,
+    validation_type: ValidationSchema,
+    annotators: List[User],
+    validators: List[User],
+    annotation_tasks: List[Task],
+    expected_output: List[User],
```
cakeinsauce, line 1325: You pass List of dicts, not User or Task

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1286,97 @@ def test_get_page_number_combinations(
...
+    with patch(
+        "annotation.distribution.main.read_user",
+        side_effect=read_user_side_effect,
+    ), patch(
+        "annotation.distribution.main.create_user",
+        side_effect=create_user_side_effect,
+    ):
+        output = prepare_users(db, users_id)
+        for a, e in zip(output, expected_output):
+            assert a.user_id == e.user_id
```
cakeinsauce, line 1382: Why dont we compare output with expected output? When we have such asserts, if they fail, it is harder to read and pinpoint what exactly failed and why

annotation/tests/test_distribution.py
UkiStefanovic: I thought that was okay cause I saw Task type defined as:
Task = Dict[str, Union[int, List[int], bool]]

annotation/tests/test_distribution.py
UkiStefanovic: The User class does not have __eq__ and so I had to check their id's manually. Is there a better way to check output==expected?s

annotation/tests/test_distribution.py
UkiStefanovic: Also, Users are for some reason unsubscriptable and you cant call user['user_id'] on a User object, so I went with dictionaries for Users as thats probably how they are stored in the database

annotation/tests/test_distribution.py
cakeinsauce: Ask Djordje, he had the same issue, he will help

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+        print("CREATE ME")
+        return User(user_id=user_id)
+
+    with patch(
+        "annotation.distribution.main.read_user",
+        side_effect=read_user_side_effect,
+    ), patch(
+        "annotation.distribution.main.create_user",
+        side_effect=create_user_side_effect,
+    ):
```
cakeinsauce, line 1384: Please, adjust your patches with side effects, you can make it much simplier without creation of these functions.

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+    ),
+    (
+        (
+            [{"user_id": "0", "overall_load": 10, "default_load": 10}],
+            10,  # all_job_pages_sum
+            10.0,  # average_job_pages
+            10,  # users_default_load
+            10,  # users_overall_load
+            1.0,  # expected_output
+            [1],  # expected_share_loads
```
cakeinsauce, line 1408: you have one parametrized case, no need to use it then

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+    output = find_users_share_loads(
+        users,
+        all_job_pages_sum,
+        average_job_pages,
+        users_default_load,
+        users_overall_load,
+    )
+    assert output == expected_output
+    for user, expected_share_load in zip(users, expected_share_loads):
+        assert user["share_load"] == expected_share_load
```
cakeinsauce, line 1430: dont use such constructions in tests, please, let's make it simplier

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+                {"pages_number": 30, "file_id": 2},
+                {"pages_number": 40, "file_id": 3},
+            ],
+            90,
+            [
+                {"pages_number": 20, "file_id": 1},
+                {"pages_number": 30, "file_id": 2},
+                {"pages_number": 40, "file_id": 3},
+            ],
+        ),
```
cakeinsauce, line 1454: you only change the actual number, let's simplify parametrize cases and construct full input/output data in the test

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+        (
+            [
+                {"pages_number": 10, "file_id": 0},
+                {"pages_number": 30, "file_id": 1},
+                {"pages_number": 0, "file_id": 2},
+            ],
+            20,
+            "a",
+            ([{"pages_number": 10, "file_id": 0}], 0),
```
cakeinsauce, line 1482: same as previous comment

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+def test_find_equal_files(
+    files: List[Dict[str, int]],
+    user_pages: int,
+    expected_output: List[Dict[str, int]],
+):
+    assert find_equal_files(files, user_pages) == expected_output
+
+
+@pytest.mark.parametrize(
+    ("files", "user_pages", "SPLIT_MULTIPAGE_DOC_setup", "expected_output"),
```
cakeinsauce, line 1466: format SPLIT_MULTIPAGE_DOC_setup to follow pep8

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+    user_pages: int,
+    SPLIT_MULTIPAGE_DOC_setup: str,
+    expected_output,
+):
+    def find_files_for_task_side_effect(files, pages_for_task):
+        lst = []
+        for file in files:
+            if file["pages_number"] and file["pages_number"] in pages_for_task:
+                lst.append(file)
+        return lst
```
cakeinsauce, line 1497: too complicated. if you do that for your patch, it is most likely due to incorrect patching. think what you could patch to avoid such complex thing

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+                mock_arg,
+                mock_arg,
+                mock_arg,
+                mock_arg,
+                mock_arg,
+                mock_arg,
+                mock_arg,
+            )
+            == expected_output
+        )
```
cakeinsauce, line 1528: what do we test here? seems like mock testing

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,238 @@ def test_get_page_number_combinations(
...
+            ["b908ccf0-6eab-4b8e-afb7-913372ddf4e5"],
+            [
+                User(user_id="b908ccf0-6eab-4b8e-afb7-913372ddf4e5"),
+                User(user_id="00d37dc4-b7da-45e5-bcb5-c9b82965351c"),
+            ],
+        ),
+    ),
+)
+def test_prepare_users(
+    users_id: List[UUID], example_db: Tuple[UUID], expected_output: List[User]
```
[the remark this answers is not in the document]
UkiStefanovic: The only error mypy is telling me is: Unexpected keyword argument "user_id" for "User". Im not sure how else to create User objects with a given user_id. Should I change User objects into dictionaries or should I do something else?

annotation/tests/test_distribution.py
UkiStefanovic: Maybe instead of the function read_user_side_effect and create_user_side_effect, I could just pass lambdas. Were you thinking of some other way?

annotation/tests/test_distribution.py
UkiStefanovic: The function prepare_response only does:
return [ task for task in distribute(...)]
I didnt really understand why we have that function, since it returns the same list like distribute(...)
I only made this test to cover the function, but it really doesnt test any functionality

annotation/annotation/models.py
```diff
@@ -200,6 +200,17 @@ class User(Base):
...
+    def __eq__(self, other) -> bool:
+        if not isinstance(other, User):
+            return NotImplemented
+
+        for column in self.__table__.columns:
+            if column.key != "_sa_instance_state" and other.__getattribute__(
+                column.key
+            ) != self.__getattribute__(column.key):
+                return False
+        return True
```
UkiStefanovic, line 212: After adding this __eq__, now i can test:
assert user1 == user2

annotation/tests/test_distribution.py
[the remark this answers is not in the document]
UkiStefanovic: Yesterday, I fixed the data types

annotation/tests/test_distribution.py
UkiStefanovic: To make sure all parts of code in the function are tested, I'll add one more test case

annotation/tests/test_distribution.py
UkiStefanovic: Now that I think about it, we don't need to patch the function _find_files_for_task_. I will remove this patch and it will make the test much simpler

UkiStefanovic: Also, I've just now realized that my test doesn't cover one edge case. I will add another test case for this test

annotation/tests/test_distribution.py
UkiStefanovic: I noticed that I needed to add another test case. Please tell me if you think I should still simplify the parametrize. The new test case makes it a bit more complicated

annotation/annotation/models.py
```diff
@@ -200,6 +200,17 @@ class User(Base):
         "Job", secondary=association_job_owner, back_populates="owners"
     )
 
+    def __eq__(self, other) -> bool:
```
cakeinsauce, line 203: add typing for arg please

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,278 @@ def test_get_page_number_combinations(
...
+        "annotation_tasks",
+        "expected_output",
+    ),
+    (
+        (
+            "",
+            ValidationSchema.cross,
+            [{"user_id": "0"}, {"user_id": "1"}],
+            [{"user_id": "0"}, {"user_id": "1"}],
+            [],
```
cakeinsauce, line 1310: you can simply specify tuple of ids and create list of dicts with user ids in the actual test, effectively making this parametrize much smaller and easier to read

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,278 @@ def test_get_page_number_combinations(
...
+)
+def test_prepare_users(
+    users_id: List[UUID], example_db: Tuple[UUID], expected_output: List[User]
+):
+    db = Mock()
+
+    with patch(
+        "annotation.distribution.main.read_user",
+        side_effect=lambda db, user_id: User(user_id=user_id)
+        if user_id in example_db
```
cakeinsauce, line 1371: in this test you dont have any case where user_id is not present in example_db. Please adjust test or better add edge cases

annotation/tests/test_distribution.py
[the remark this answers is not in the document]
cakeinsauce: `Tuple[UUID, ...]`

I believe I have already shared it, but once again
[link](https://docs.python.org/3/library/typing.html#:~:text=To%20denote%20a%20tuple%20which%20could%20be%20of%20any%20length%2C%20and%20in%20which%20all%20elements%20are%20of%20the%20same%20type%20T%2C%20use%20tuple%5BT%2C%20...%5D.%20To%20denote%20an%20empty%20tuple%2C%20use%20tuple%5B()%5D.%20Using%20plain%20tuple%20as%20an%20annotation%20is%20equivalent%20to%20using%20tuple%5BAny%2C%20...%5D%3A)

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,278 @@ def test_get_page_number_combinations(
...
+        "expected_users",
+    ),
+    (
+        (
+            [{"user_id": "0", "overall_load": 10, "default_load": 10}],
+            10,  # all_job_pages_sum
+            10.0,  # average_job_pages
+            10,  # users_default_load
+            10,  # users_overall_load
+            1.0,  # expected_output
```
cakeinsauce, line 1398: you dont need such comments for params you have already denoted

cakeinsauce: Same applies for comments below

annotation/tests/test_distribution.py
cakeinsauce: I did not get the problem here. You can only provide numbers, it wont complicate the test

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1291,278 @@ def test_get_page_number_combinations(
...
+            "",
+            ([], 0),
+        ),
+    ),
+)
+def test_find_small_files(
+    files: List[Dict[str, int]],
+    user_pages: int,
+    split_multipage_doc_setup: str,
+    expected_output,
```
cakeinsauce, line 1540: typing, please

annotation/tests/test_distribution.py
UkiStefanovic: Im sorry about that. My intention was to cover when user_id is not in example_db with the second test case. I guess in the meantime I copy-pasted the lists from above.

annotation/tests/test_distribution.py
UkiStefanovic: Mocked the SPLIT_MULTIPAGE_DOC var and simplified the test

annotation/tests/test_distribution.py
UkiStefanovic: Sorry again. Now I know what you meant by simplify the parametrize

annotation/annotation/models.py
```diff
@@ -200,6 +200,17 @@ class User(Base):
...
     )
 
+    def __eq__(self: "User", other: Any) -> bool:
+        if not isinstance(other, User):
+            return NotImplemented
+
+        for column in self.__table__.columns:
+            if column.key != "_sa_instance_state" and other.__getattribute__(
+                column.key
+            ) != self.__getattribute__(column.key):
```
cakeinsauce, line 210:
```suggestion
            ) != other.__getattribute__(column.key):
```

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1290,285 @@ def test_get_page_number_combinations(
...
+            "",
+            ValidationSchema.cross,
+            (0, 1),
+            [],
+        ),
+        (
+            "1",
+            ValidationSchema.hierarchical,
+            1,
+            [2],
```
cakeinsauce, line 1313: list -> tuple

annotation/annotation/models.py
UkiStefanovic: How did I not notice that... Sorry

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1290,285 @@ def test_get_page_number_combinations(
...
+    files_page_count: List[int],
+    user_pages: int,
+    expected_output: List[Dict[str, int]],
+):
+    file_count = len(files_page_count)
+    files = [
+        {"pages_number": x, "file_id": i}
+        for x, i in zip(files_page_count, range(file_count))
+    ]
+    print(files)
```
cakeinsauce, line 1511: remove print please

annotation/tests/test_distribution.py
cakeinsauce: it is still the same
image

annotation/tests/test_distribution.py
cakeinsauce: still the same
image

annotation/tests/test_distribution.py
```diff
@@ -1277,3 +1290,285 @@ def test_get_page_number_combinations(
...
+        split_multipage_doc_setup,
+    ):
+        output = choose_validators_users(
+            validation_type,
+            annotators,
+            validators,
+            annotation_tasks=(
+                [{"user_id": user_id} for user_id in annotation_tasks_ids]
+                if isinstance(annotation_tasks_ids, Iterable)
+                else [{"user_id": annotation_tasks_ids}]
```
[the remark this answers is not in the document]
UkiStefanovic: Yeah, I've only now noticed that (1) is an int, but (1,) is a tuple with one element

annotation/annotation/models.py
```diff
@@ -200,6 +200,17 @@ class User(Base):
         "Job", secondary=association_job_owner, back_populates="owners"
     )
 
+    def __eq__(self: "User", other: object) -> bool:
```
cakeinsauce, line 203: never type self, excessive

annotation/tests/test_distribution.py
UkiStefanovic: ⁠Regarding this, I thought that i cant remove file_id's. If we have multiple files with the same page_number, then we should keep file_id's to check if the output was as expected. But if i want to just see if the tests work correctly, i could just:

`​assert sum([file["pages_number"] for file in output]) == user_pages`

In that case, file_id's are redundant and I'll remove them completely

UkiStefanovic: Nevermind that, I think I got it

annotation/annotation/models.py
cakeinsauce: you have to have self in any method, but you shouldn't specify typing for self, it is obvious the same as the class in which this method is located

cakeinsauce (approved):
Great job!

PR #895, opened 2024-08-11
test: Add parameterized test cases for validate_user_actions function

BogicevicMilan:
- Added parameterized test cases to validate various scenarios for the validate_user_actions function.
- Included cases for missing parameters, invalid user actions, and handling unprocessed pages.


annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    (
+        True,
+        {1},
+        {2},
+        {3},
+        True,
+        True,
+        "Cannot finish validation task. There are not processed pages: " "{3}",
+    ),
+)
```
djo753, line 90: Same as the previous comment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@pytest.fixture(scope="function")
```
djo753, line 93: Scope is "function" by default. No need to explicitly write it.
```suggestion
@pytest.fixture()
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+def db_session():
+    return Mock(spec=Session)
```
djo753, line 95: Use **yield** instead of return inside fixtures.
```suggestion
    yield Mock(spec=Session)
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@pytest.fixture(scope="function")
```
djo753, line 98: Same as the previous comments. "function" is default scope

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+def task_info():
+    return {
```
djo753, line 100:
```suggestion
    yield {
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+@pytest.mark.parametrize(
+    ("validation_type", "is_validation"),
+    (
+        (ValidationSchema.validation_only, True),
+        ("other_type", False),
+        ("other_type", True),
+    ),
+)
+def test_validate_task_info_positive_case(
```
djo753, line 115: No need for 'positive_case'. If there is no description, it is considered a success test case.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    db_session: Session,
+    task_info: dict,
```
djo753, line 117: Add types to dict -> Dict[x, x] 
Dict is from `typing`

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+            db_session, task_info, validation_type
+        )
+        mock_validate_files_info.assert_called_once_with(db_session, task_info)
+
+
+@pytest.mark.parametrize(
+    ("validation_type", "is_validation"),
+    ((ValidationSchema.validation_only, False),),
+)
+def test_validate_task_info_negative_case(
```
djo753, line 139: Instead of 'negative_case' we should write a reason or description of why the test should raise an exception. For example , here we could call it 'non_exclusive_validation_info'

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    "validation_type, is_validation, mock_function, mock_return_value",
+    (
+        (ValidationSchema.cross, True, "check_cross_annotating_pages", True),
+        (ValidationSchema.validation_only, True, None, True),
+        (ValidationSchema.validation_only, False, None, True),
+    ),
+)
+def test_validate_users_info_positive_case(
+    db_session: Session,
+    task_info: dict,
```
djo753, line 161: Add type hint for keys and values of dict

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    else:
+        db_session.query().filter_by().first.return_value = mock_return_value
+        validate_users_info(db_session, task_info, validation_type)
+
+
+@pytest.mark.parametrize(
+    ("is_validation", "expected_error_message"),
+    (
+        (True, "user 1 is not assigned as validator for job 2"),
+        (False, "user 1 is not assigned as annotator for job 2"),
```
djo753, line 182: Use f"{task_info.job_id}" rather than hard-coding the error message. It would make the code more readable.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    ),
+)
+def test_validate_users_info_negative_case(
+    db_session: Session,
+    task_info: dict,
```
djo753, line 187: Same as other dict

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@pytest.mark.parametrize(
+    ("is_validation", "expected_error_message"),
+    (
+        (True, "user 1 is not assigned as validator for job 2"),
+        (False, "user 1 is not assigned as annotator for job 2"),
+    ),
+)
+def test_validate_users_info_negative_case(
```
djo753, line 185: Same as the other negative case, we should think of a better description for it.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+        "file_id": 1,
+        "job_id": 2,
+        "pages": [1, 2, 3],
+    }
+
+    mock_file = Mock(spec=File)
+    mock_file.pages_number = 3
+
+    mock_query = db_session.query.return_value
+    mock_query.filter_by.return_value.first.return_value = mock_file
```
djo753, line 211:
```suggestion
    db_session.query().filter().first() = mock_file
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    task_info["is_validation"] = is_validation
+    db_session.query().filter_by().first.return_value = None
+    with pytest.raises(FieldConstraintError) as excinfo:
+        validate_users_info(
+            db_session, task_info, ValidationSchema.validation_only
+        )
+    assert expected_error_message in str(excinfo.value)
+
+
+def test_validate_files_info_positive_case(db_session: Session):
```
djo753, line 200: Same as previous, no need for 'positive_case'

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    validation_type: ValidationSchema,
+    is_validation: bool,
+):
+    task_info["is_validation"] = is_validation
+
+    with patch(
+        "annotation.tasks.services.validate_users_info"
+    ) as mock_validate_users_info, patch(
+        "annotation.tasks.services.validate_files_info"
+    ) as mock_validate_files_info:
```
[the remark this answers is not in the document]
BogicevicMilan: Thanks.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+        ]
+    )
+    mock_query.filter_by.assert_called_once_with(file_id=1, job_id=2)
+
+
+@pytest.mark.parametrize(
+    ("task_info", "mock_file", "expected_error_message"),
+    TEST_PARAMS_INVALID_FILES,
+)
+def test_validate_files_info_negative_case(
```
djo753, line 228: 'negative_case' -> description of failure

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    db_session: Session,
+    task_info: dict,
```
djo753, line 230: type hints

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    mock_file: Mock,
+    expected_error_message: str,
+):
+    db_session.query().filter_by().first.return_value = mock_file
```
djo753, line 234: Good usage!

BogicevicMilan: Thanks.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+    with pytest.raises(FieldConstraintError) as excinfo:
+        validate_files_info(db_session, task_info)
+
+    assert expected_error_message in str(excinfo.value)
```
djo753, line 239: We should consider 2 negative test cases here. Since they occur under different circumstances. This should help with the parametrize (there would be no need for it anymore) and with it, no need for the global variable.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+    validate_files_info(db_session, task_info)
+
+    assert db_session.query.call_count == 1
+    db_session.query.assert_has_calls(
+        [
+            call(File),
+        ]
+    )
+    mock_query.filter_by.assert_called_once_with(file_id=1, job_id=2)
```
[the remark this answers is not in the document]
BogicevicMilan: The variable `mock_query` is used later in the test to ensure that query operations are correctly mocked.

annotation/tests/test_tasks_services.py
BogicevicMilan: The variable `mock_query` is used later in the test to ensure that query operations are correctly mocked.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+    assert expected_error_message in str(excinfo.value)
+
+
+def test_check_cross_annotating_pages_positive_case(db_session: Session):
+    task_info = {"user_id": 1, "file_id": 2, "job_id": 3, "pages": {4, 5}}
+    existing_pages = []
+
+    mock_query = db_session.query.return_value
+    mock_query.filter.return_value.all.return_value = [(existing_pages,)]
```
djo753, line 247: We should consider the idea from the previous comment to simply 
mock_query -> db.query().

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+        assert db_session.query.call_count == 1
+        db_session.query.assert_has_calls(
+            [
+                call(ManualAnnotationTask.pages),
+            ]
+        )
+        mock_query.filter.assert_called_once()
+
+    except FieldConstraintError:
+        pytest.fail("FieldConstraintError was raised unexpectedly")
```
djo753, line 217: Are we expecting a successful unit test to fail? If not, we should remove this `try: except:` block

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+def test_check_cross_annotating_pages_negative_case(db_session: Session):
```
djo753: We should describe the negative case

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    task_info = {"user_id": 1, "file_id": 2, "job_id": 3, "pages": {4, 5}}
+    existing_pages = [4, 5]
+
+    mock_query = db_session.query.return_value
+    mock_query.filter.return_value.all.return_value = [(existing_pages,)]
+
+    expected_error_message = (
+        "within cross validation job user can't validate file's pages "
+        "that are already distributed in annotation tasks for this user: "
+        "{4, 5}"
```
djo753, line 234: We should utilize f string in situations like these

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,311 @@
...
+    with pytest.raises(HTTPException) as excinfo:
+        validate_user_actions(
+            is_validation=is_validation,
+            failed=failed,
+            annotated=annotated,
+            not_processed=not_processed,
+            annotation_user=annotation_user,
+            validation_user=validation_user,
+        )
+    assert excinfo.value.detail == expected_error_message
```
djo753: Nicely done!

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+from annotation.tasks.services import (
+    check_cross_annotating_pages,
+    validate_files_info,
+    validate_task_info,
+    validate_user_actions,
+    validate_users_info,
+)
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 19: You can simply use it inside the test as a patch - as you do with other patches

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+        "annotation.tasks.services.validate_files_info"
+    ) as mock_validate_files_info:
+        validate_task_info(db_session, task_info, validation_type)
+        mock_validate_users_info.assert_called_once_with(
+            db_session, task_info, validation_type
+        )
+        mock_validate_files_info.assert_called_once_with(db_session, task_info)
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 48: same as previous comment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+):
+    db_session = mock_session()
+    task_info = {"is_validation": False}
+    validation_type = ValidationSchema.validation_only
+
+    with pytest.raises(FieldConstraintError):
+        validate_task_info(db_session, task_info, validation_type)
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 60: same as previous comment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+@pytest.mark.parametrize(
+    ("validation_type", "is_validation"),
+    (
+        (ValidationSchema.validation_only, True),
+        (ValidationSchema.validation_only, False),
```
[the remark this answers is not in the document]
cakeinsauce: Additionally, you already have `validation_type` variable in the code later with basically the same assignment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+
+    validation_type = ValidationSchema.validation_only
+    db_session.query().filter_by().first().return_value = True
+
+    validate_users_info(db_session, task_info, validation_type)
+
+    assert db_session.query.call_count == 2
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 86: same as previous comment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+def test_validate_users_info_calls_check_cross_annotating_pages(
```
cakeinsauce, line 87: lets keep naming simpler, e.g. test_validate_users_info_cross_validation

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    with patch(
+        f"annotation.tasks.services.{'check_cross_annotating_pages'}"
+    ) as mock_func:
+        validate_users_info(db_session, task_info, ValidationSchema.cross)
+        mock_func.assert_called_once_with(db_session, task_info)
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 104: same as previous comment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+@pytest.mark.parametrize(
+    ("is_validation", "validator_or_annotator"),
+    [
+        (True, "validator"),
+        (False, "annotator"),
+    ],
```
cakeinsauce, line 110: list -> tuple

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+        f"user 1 is not assigned as {validator_or_annotator} for job 2"
+    )
+
+    with pytest.raises(FieldConstraintError, match=expected_error_message):
+        validate_users_info(
+            db_session, task_info, ValidationSchema.validation_only
+        )
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 135: same as previous comment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+    mock_query.filter_by.return_value.first.return_value = mock_file
+
+    validate_files_info(db_session, task_info)
+
+    assert db_session.query.call_count == 1
+    db_session.query.assert_has_calls((call(File),))
+    mock_query.filter_by.assert_called_once_with(file_id=1, job_id=2)
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 157: same as previous comment

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+            r"pages \(\{4\}\) do not belong to file",
+        ),
+        (
+            {
+                "file_id": 1,
+                "job_id": 2,
+                "pages": [1, 2, 3],
+            },
+            None,
+            r"file with id 1 is not assigned for job 2",
```
cakeinsauce, line 177: Let's split these cases into two tests. It will simplify this a lot

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+    db_session = mock_session()
+    db_session.query().filter_by().first.return_value = mock_file
+
+    with pytest.raises(
+        FieldConstraintError, match=expected_error_message_regex
+    ):
+        validate_files_info(db_session, task_info)
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
cakeinsauce, line 196: same as previous comment

annotation/tests/test_tasks_services.py
cakeinsauce: agree, we definitely don't want to have try..except within a test

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+                "file_id": 1,
+                "job_id": 2,
+                "pages": [1, 2, 3],
+            },
+            None,
+            r"file with id 1 is not assigned for job 2",
+        ),
+    ),
+)
+def test_validate_files_info_raises_error_for_invalid_file_info(
```
cakeinsauce, line 181: e.g. test_validate_files_info_invalid_file

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+            ]
+        )
+        mock_query.filter.assert_called_once()
+
+    except FieldConstraintError:
+        pytest.fail("FieldConstraintError was raised unexpectedly")
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
+def test_check_cross_annotating_pages_raise_error_when_page_already_annotated(
```
cakeinsauce, line 221: e.g. test_check_cross_annotating_pages_annotated

annotation/tests/test_tasks_services.py
cakeinsauce: Addition from me - you shouldn't validate the whole message for a match - just check important parts, like type of validation, user ids, etc.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+        (
+            True,
+            {1},
+            set(),
+            set(),
+            False,
+            True,
+            "Missing `annotation_user_for_failed_pages` param "
+            "for failed pages.",
+        ),
```
cakeinsauce, line 261: For this and other test cases: 
1. is_validation is always true - remove it
2. no need to check for the whole error message - you can match important parts which are different from each other. And simply looking at this I would split this test into multiple test cases (but only if they are logically separable)

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+            {2},
+            {3},
+            True,
+            True,
+            "Cannot finish validation task. There are not processed pages: "
+            "{3}",
+        ),
+    ),
+)
+def test_validate_user_actions_raises_error_for_invalid_user_states(
```
cakeinsauce, line 304: e.g. test_validate_user_actions_invalid

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+    with pytest.raises(HTTPException) as excinfo:
+        validate_user_actions(
+            is_validation=is_validation,
+            failed=failed,
+            annotated=annotated,
+            not_processed=not_processed,
+            annotation_user=annotation_user,
+            validation_user=validation_user,
+        )
+    assert excinfo.value.detail == expected_error_message
```
cakeinsauce, line 322: https://docs.pytest.org/en/stable/how-to/assert.html#matching-exception-messages

annotation/tests/test_tasks_services.py
BogicevicMilan: Done

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok:)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok:)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: fixed

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok;)

annotation/tests/test_tasks_services.py
BogicevicMilan: fixed

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,322 @@
...
+                call(ManualAnnotationTask.pages),
+            ]
+        )
+        mock_query.filter.assert_called_once()
+
+    except FieldConstraintError:
+        pytest.fail("FieldConstraintError was raised unexpectedly")
+
+
+@patch("sqlalchemy.orm.Session", spec=True)
```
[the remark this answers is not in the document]
BogicevicMilan: Ok.

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok.

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok.

BogicevicMilan: I attempted to use match within pytest.raises for testing error messages, but since the error description is within the detail attribute of the exception, I think it's more accurate to directly assert excinfo.value.detail against the expected message.

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,321 @@
...
+        with pytest.raises(FieldConstraintError):
+            validate_task_info(db_session, task_info, validation_type)
+
+
+@pytest.mark.parametrize(
+    ("is_validation"),
+    (
+        (True),
+        (False),
+    ),
```
cakeinsauce, line 63:
```suggestion
    "is_validation", (True, False)
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,321 @@
...
+)
```
cakeinsauce, line 64:
```suggestion
@pytest.mark.parametrize("is_validation", (True, False))
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,321 @@
...
+        validate_users_info(db_session, task_info, ValidationSchema.cross)
+        mock_func.assert_called_once_with(db_session, task_info)
+
+
+@pytest.mark.parametrize(
+    ("is_validation", "validator_or_annotator"),
+    (
+        (True, "validator"),
+        (False, "annotator"),
+    ),
```
cakeinsauce, line 103:
```suggestion
    ((True, "validator"), (False, "annotator")),
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,321 @@
...
+    validation_type: ValidationSchema,
+    is_validation: bool,
+):
+    with patch("sqlalchemy.orm.Session", spec=True) as mock_session, patch(
+        "annotation.tasks.services.validate_users_info"
+    ) as mock_validate_users_info, patch(
+        "annotation.tasks.services.validate_files_info"
+    ) as mock_validate_files_info:
+
+        db_session = mock_session()
```
cakeinsauce, line 37: Could you elaborate why do that? why not simply use mock_session everywhere?

annotation/tests/test_tasks_services.py
cakeinsauce: I did not get the problem. Could you provide screenshot/examples why match wouldn't work here?

annotation/tests/test_tasks_services.py
BogicevicMilan: fixed;)

annotation/tests/test_tasks_services.py
BogicevicMilan: fixed:)

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,321 @@
...
+        "validation_user",
+        "expected_error_message_pattern",
+    ),
+    (
+        (
+            set(),
+            set(),
+            set(),
+            True,
+            True,
```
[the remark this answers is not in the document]
BogicevicMilan: fixed;)

annotation/tests/test_tasks_services.py
BogicevicMilan: Since I've mocked these two functions, there's no need to use db_session here, as those functions won't be interacting with it. Therefore, I'll remove that part and use None instead.

annotation/tests/test_tasks_services.py
cakeinsauce: still the same

annotation/tests/test_tasks_services.py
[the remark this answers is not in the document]
cakeinsauce: still here

annotation/tests/test_tasks_services.py
cakeinsauce: still the same

annotation/tests/test_tasks_services.py
cakeinsauce: Didnt get the answer and the code is still the same

annotation/tests/test_tasks_services.py
BogicevicMilan: I usually fix the issue locally first, then I write 'ok,' 'fixed,' or 'done' once it's resolved before pushing the changes. Sorry about that!

annotation/tests/test_tasks_services.py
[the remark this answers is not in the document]
BogicevicMilan: I usually fix the issue locally first, then I write 'ok,' 'fixed,' or 'done' once it's resolved before pushing the changes. Sorry about that!

annotation/tests/test_tasks_services.py
BogicevicMilan: I reviewed the issue and found that the reason for the AssertionError with the message Regex pattern '.*Cannot finish validation task. There are not processed pages.*' does not match '' is that the error message is stored in the detail attribute of the HTTPException.
The match parameter in pytest.raises does not compare the detail attribute directly; instead, it attempts to match the entire exception message. Since the detail field is where the message is stored, match ends up comparing against an empty string, which leads to the failure.

annotation/tests/test_tasks_services.py
BogicevicMilan: I've made the changes as it seemed more appropriate to write two mock objects for the functions being called. This approach felt more readable and understandable to me, especially since db_session is used for database connection, and I didn't want to add extra functionality. I was aiming to adhere to the single responsibility principle.

annotation/tests/test_tasks_services.py
BogicevicMilan: Now it is changed.

PR #896, opened 2024-08-13
test(annotation): cover construct_annotated_doc

djo753:


annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
...
+                annotated_doc.file_id,
+                db,
+                None,
+            )
+
+
+def test_construct_annotated_doc_db_error(
+    annotated_doc,
+    annotation_doc_for_save,
+):
```
BogicevicMilan, line 383: Add types, please

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
...
+@pytest.mark.parametrize(
+    ("is_latest", "new_equal_latest"),
+    ((True, False), (True, True)),  # Success path  # return latest_doc
+)
+def test_construct_annotated_doc(
+    annotated_doc,
+    annotation_doc_for_save,
+    is_latest,
+    new_equal_latest,
+):
```
BogicevicMilan, line 288: Add types, please

annotation/tests/test_annotations.py
```diff
@@ -107,6 +112,40 @@ def annotation_manifest():
     }
 
 
+@pytest.fixture
+def annotation_doc_for_save():
```
BogicevicMilan, line 116: If you're using this function in only one place, there's no need to create a fixture.

annotation/tests/test_annotations.py
```diff
@@ -53,7 +58,7 @@ def annotator():
 
 
 @pytest.fixture
-def annotation_job(annotator: User, annotation_file: File, categories: Job):
+def annotation_job(annotator, annotation_file, categories):
```
BogicevicMilan, line 61: If you're using this function in only one place, there's no need to create a fixture.

BogicevicMilan: Add types.

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
             annotated_doc.tenant,
             moto_s3,
         )
+
+
+@pytest.mark.parametrize(
+    ("is_latest", "new_equal_latest"),
+    ((True, False), (True, True)),  # Success path  # return latest_doc
```
BogicevicMilan, line 281: `is_latest` is True in both cases, so you can remove it.

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
...
+    pages_sha_latest = ({"1": "a"}, "b")
+    pages_sha_non_latest = ({"1": "a"}, "c")
+
+    db = Mock()
+
+    with patch.object(db, "add") as mock_db_add, patch.object(
+        db, "add_all"
+    ) as mock_db_add_all, patch.object(
+        db,
+        "commit",
```
BogicevicMilan, line 306: You can add methods in db  mock instead of using patch several times.

annotation/tests/test_annotations.py
djo753: It is used within another function as a parameter. I will leave it as a fixture.

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
...
+                doc.failed_validation_pages,
+                doc.user,
+            )
+        else:
+            mock_get_pages_sha.assert_called_once_with(
+                doc.pages,
+                latest_doc.revision,
+                doc.validated,
+                doc.failed_validation_pages,
+                doc.user,
```
BogicevicMilan, line 351: Remove logic from the tests. Conditional if statements are often the most common source of errors in tests. It might be beneficial to split this function into several parts to improve clarity and reduce potential issues.

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
...
+            )
+        mock_check_docs.assert_called_once()
+        if not new_equal_latest:
```
BogicevicMilan, line 354: same as previous comment

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
...
+                annotated_doc.user,
+                None,
+                annotated_doc.job_id,
+                annotated_doc.file_id,
+                doc,
+                TEST_TENANT,
+                s3_file_path,
+                s3_file_bucket,
+                latest_doc,
+                1,  # task_id
```
BogicevicMilan, line 425: Please note that the task_id is known according to the function signature, so the comment explaining it can be omitted.

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,156 @@ def test_create_manifest_json(
...
+            annotated_doc.file_id,
+            doc,
+            TEST_TENANT,
+            s3_file_path,
+            s3_file_bucket,
+            latest_doc,
+            1,  # task_id
+            is_latest,
+        )
+        if is_latest:
```
BogicevicMilan, line 337: Remove logic from the test.

annotation/tests/test_annotations.py
djo753:
```suggestion
    db.add() = Mock()
    db.add_all() = Mock()
```
Would you say this is more readable? Should we do this for every db method?

annotation/tests/test_annotations.py
djo753: Separated it into 2 tests

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+def test_construct_annotated_doc(
+    annotated_doc: AnnotatedDoc,
+    annotation_doc_for_save: DocForSaveSchema,
+):
+    doc = annotation_doc_for_save
+    latest_doc = annotated_doc
+    is_latest = False
+    new_equal_latest = False
+    s3_path = (
+        f"{S3_START_PATH}/{annotated_doc.job_id}/" f"{annotated_doc.file_id}"
```
cakeinsauce, line 288:
```suggestion
        f"{S3_START_PATH}/{annotated_doc.job_id}/{annotated_doc.file_id}"
```

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+
+
+def test_construct_annotated_doc(
+    annotated_doc: AnnotatedDoc,
+    annotation_doc_for_save: DocForSaveSchema,
+):
+    doc = annotation_doc_for_save
+    latest_doc = annotated_doc
+    is_latest = False
+    new_equal_latest = False
```
cakeinsauce, line 286: you use them only once, lets remove unnecessary vars. for the clarity you should use kw args in the function you utilize these vars

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+        "annotation.annotations.main.construct_document_links", return_value=[]
+    ) as mock_construct_doc_links, patch(
+        "annotation.annotations.main.convert_bucket_name_if_s3prefix",
+        return_value=TEST_TENANT,
+    ) as mock_convert_bucket, patch(
+        "annotation.annotations.main.upload_pages_to_minio"
+    ) as mock_upload_pages, patch(
+        "annotation.annotations.main.create_manifest_json"
+    ) as mock_create_manifest:
+        construct_annotated_doc(
```
cakeinsauce, line 316: construct_annotated_doc returns annotated doc. You should definitely check the output of the function. If we just assert mocks, it becomes useless with mock testing

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+
+
+def test_construct_annotated_doc_new_equals_latest(
+    annotated_doc: AnnotatedDoc,
+    annotation_doc_for_save: DocForSaveSchema,
+):
+    doc = annotation_doc_for_save
+    latest_doc = annotated_doc
+    is_latest = True
+    new_equal_latest = True
```
cakeinsauce, line 372: same as previous comment

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+    db = Mock()
+
+    with patch(
+        "annotation.annotations.main.get_pages_sha",
+        return_value=pages_sha_latest,
+    ) as mock_get_pages_sha, patch(
+        "annotation.annotations.main.check_docs_identity",
+        return_value=new_equal_latest,
+    ) as mock_check_docs:
+        construct_annotated_doc(
```
cakeinsauce, line 388: same as previous comment - check for the output

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+    pages_sha_latest = ({"1": "a"}, "b")
+    pages_sha_non_latest = ({"1": "a"}, "c")
+
+    db = Mock()
+    db.add = Mock()
+    db.add_all = Mock()
+    db.commit = Mock(
+        side_effect=IntegrityError(
+            statement="TEST", params=("testuser",), orig=Exception("TESTING")
+        )
```
cakeinsauce, line 433: is it required to have these params in the exception? if not, lets omit it

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+    )
+    db.rollback = Mock()
```
cakeinsauce, line 435: same as previous comment

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+    with patch(
+        "annotation.annotations.main.get_pages_sha",
+        return_value=pages_sha_latest if is_latest else pages_sha_non_latest,
```
cakeinsauce, line 438: again, not the best practice to put logic into a test. either split into two different tests or rethink this part

annotation/tests/test_annotations.py
```diff
@@ -235,3 +274,188 @@ def test_create_manifest_json(
...
+        mock_convert_bucket.assert_called_once_with(TEST_TENANT)
+        mock_upload_pages.assert_called_once_with(
+            pages=doc.pages,
+            pages_sha=pages_sha_non_latest[0],
+            s3_path=s3_path,
+            bucket_name=TEST_TENANT,
+            s3_resource=None,
+        )
+        mock_create_manifest.assert_called_once_with(
+            ANY,
```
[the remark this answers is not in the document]
djo753: Here we encounter the same problem as with the asserting the actual output of the function. Since assert compares object **identity** and not object attributes. I would need to define a `__eq__ `method inside of `AnnotatedDoc `model for it to be able to work.  Do we have any better ideas than that?

annotation/tests/test_annotations.py
djo753: Added __eq__ method to AnnotatedDoc.

cakeinsauce (approved):
LGTM!

PR #897, opened 2024-08-13
test(annotation): cover upload json and pages to minio

djo753:


annotation/tests/test_annotations.py
```diff
@@ -235,3 +276,219 @@ def test_create_manifest_json(
...
+                s3_file_bucket,
+                latest_doc,
+                1,
+                is_latest,
+            )
+            assert db.commit.assert_called_once()
+            assert db.rollback.assert_called_once()
+
+
+def test_upload_json_to_minio():
```
BogicevicMilan, line 466: Looks good.

annotation/tests/test_annotations.py
```diff
@@ -235,3 +276,219 @@ def test_create_manifest_json(
...
+        mock_get_storage.return_value.upload_obj = Mock()
+
+        upload_json_to_minio(test_json, path_to_object, bucket_name, None)
+        mock_get_storage.assert_called_once_with(bucket_name)
+        mock_get_storage().upload_obj.assert_called_once_with(
+            target_path=path_to_object, file=ANY
+        )
+
+
+def test_upload_pages_to_minio(moto_s3: boto3.resource):
```
BogicevicMilan, line 483: Looks good.

annotation/tests/test_annotations.py
```diff
@@ -482,3 +484,34 @@ def test_construct_annotated_doc_db_error(
...
+
+    with patch(
+        "annotation.annotations.main.bd_storage.get_storage"
+    ) as mock_get_storage:
+        mock_get_storage.return_value.upload_obj = Mock()
+
+        upload_json_to_minio(test_json, path_to_object, bucket_name, None)
+        mock_get_storage.assert_called_once_with(bucket_name)
+        mock_get_storage().upload_obj.assert_called_once_with(
+            target_path=path_to_object, file=ANY
```
cakeinsauce, line 502: As I see file is not ANY, it is a byte stream. Could you change this?

annotation/tests/test_annotations.py
```diff
@@ -482,3 +484,34 @@ def test_construct_annotated_doc_db_error(
...
+    test_json = json.dumps({"test": 1})
+    path_to_object = "path"
+    bucket_name = TEST_TENANT
+
+    with patch(
+        "annotation.annotations.main.bd_storage.get_storage"
+    ) as mock_get_storage:
+        mock_get_storage.return_value.upload_obj = Mock()
+
+        upload_json_to_minio(test_json, path_to_object, bucket_name, None)
```
cakeinsauce, line 499: s3_resource is not optional. Pass mock if you dont need it

annotation/tests/test_annotations.py
```diff
@@ -482,3 +484,34 @@ def test_construct_annotated_doc_db_error(
...
+
+def test_upload_json_to_minio():
+    test_json = json.dumps({"test": 1})
+    path_to_object = "path"
+    bucket_name = TEST_TENANT
+
+    with patch(
+        "annotation.annotations.main.bd_storage.get_storage"
+    ) as mock_get_storage:
+        mock_get_storage.return_value.upload_obj = Mock()
```
cakeinsauce, line 497: FYI you can patch it in a single line, will be much cleaner

annotation/tests/test_annotations.py
djo753: The function doesn't utilize s3_resource anymore (because of bd_storage). Will pass in Mock()

annotation/tests/test_annotations.py
djo753: `assert_called_once_with` is comparing the identity of the byte stream, that is the reason for ANY.
```suggestion
            target_path=path_to_object, =io.BytesIO(test_json.encode("UTF-8"))
```
This results in an error:
```
E           AssertionError: expected call not found.
E           Expected: upload_obj(target_path='path', file=<_io.BytesIO object at 0x070DEE60>)
E           Actual: upload_obj(target_path='path', file=<_io.BytesIO object at 0x070EEA78>)
```

PR #898, opened 2024-08-13
test(annotation): cover update_pages_array and check_if_kafka_message_is_needed"

djo753:


annotation/tests/test_annotations.py
```diff
@@ -235,3 +278,253 @@ def test_create_manifest_json(
...
+    db.commit = Mock()
+    check_if_kafka_message_is_needed(
+        db,
+        annotated_doc,
+        None,
+        annotated_doc.job_id,
+        annotated_doc.file_id,
+        TEST_TENANT
+    )
+    db.commit.assert_called_once()
```
BogicevicMilan, line 530: Add new line here.

annotation/tests/test_annotations.py
```diff
@@ -235,3 +278,253 @@ def test_create_manifest_json(
...
+    resulting_set = update_pages_array(
+        {1, 2, 3},
+        {4},
+        {1,2}
+    )
+    assert resulting_set == {3, 4}
+
+
+def test_kafka_message_needed(annotated_doc: AnnotatedDoc):
+    db = Mock()
```
BogicevicMilan, line 508: In this test, there is no assertion or error handling, which means that this is not a proper test. It needs to be corrected.

annotation/tests/test_annotations.py
BogicevicMilan: It looks like your pre-commit hook isn't working; otherwise, it wouldn't have allowed you to commit such code.

annotation/tests/test_annotations.py
```diff
@@ -513,3 +515,35 @@ def test_upload_pages_to_minio(moto_s3: boto3.resource):
...
+    check_if_kafka_message_is_needed(
+        db,
+        annotated_doc,
+        annotated_doc,
+        annotated_doc.job_id,
+        annotated_doc.file_id,
+        TEST_TENANT,
+    )
+    db.assert_not_called
```
cakeinsauce, line 535: could you check for the result/raised exceptions? additionally, assert_not_called is a method

djo753: The function doesn't explicitly state that it raises an exception. 
I can check for the error raised by db.commit(), however, that would just be testing the side_effect of a Mock() object.
Additionally, the return value of the function is None, so I don't have any result to assert.

cakeinsauce (approved):
Great!

PR #899, opened 2024-08-13
test(tasks): test create_annotation_task function

BogicevicMilan:


annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+
+def test_create_annotation_task_success():
+
+    with patch("sqlalchemy.orm.Session", spec=True) as mock_session, patch(
+        "annotation.jobs.services.update_user_overall_load"
+    ) as mock_update_user_overall_load, patch(
+        "annotation.models.ManualAnnotationTask"
+    ), patch(
+        "annotation.schemas.tasks.ManualAnnotationTaskInSchema"
+    ):
```
djo753, line 311: No need to mock models

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+            failed=failed,
+            annotated=annotated,
+            not_processed=not_processed,
+            annotation_user=True,
+            validation_user=True,
+        )
+    assert re.match(expected_error_message_pattern, excinfo.value.detail)
+
+
+def test_create_annotation_task_success():
```
djo753, line 303: We don't need to write success, it is already considered unless another description is written

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+
+    with patch("sqlalchemy.orm.Session", spec=True) as mock_session, patch(
+        "annotation.jobs.services.update_user_overall_load"
+    ) as mock_update_user_overall_load, patch(
+        "annotation.models.ManualAnnotationTask"
+    ), patch(
+        "annotation.schemas.tasks.ManualAnnotationTaskInSchema"
+    ):
+
+        db_session = mock_session()
```
djo753, line 313: Do we need db_session? We already have a mock_session and can utilize it in asserts

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+        create_annotation_task(
+            db_session,
```
djo753, line 315:
```suggestion
            mock_session,
```

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+            annotation_user=True,
+            validation_user=True,
+        )
+    assert re.match(expected_error_message_pattern, excinfo.value.detail)
+
+
+def test_create_annotation_task_success():
+
+    with patch("sqlalchemy.orm.Session", spec=True) as mock_session, patch(
+        "annotation.jobs.services.update_user_overall_load"
```
djo753, line 306:
```suggestion
        "annotation.tasks.services.update_user_overall_load"
```
Wrong patch, this error creates bad assertions.

annotation/tests/test_tasks_services.py
BogicevicMilan: fixed

annotation/tests/test_tasks_services.py
BogicevicMilan: fixed

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+                file_id=1,
+                pages={1, 2},
+                job_id=2,
+                user_id=uuid4(),
+                is_validation=True,
+                deadline=None,
+            ),
+        )
+
+        assert db_session.add.call_count == 2
```
djo753, line 326:
```suggestion
        assert db_session.add.call_count == 1
```
Because you did not mock the `update_user_overall_load` , it is being called and and db.add is being called twice. With the correct patch of said function, add is begin called only once.

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+        mock_update_user_overall_load.assert_not_called()
```
djo753, line 327: For the same reason as the previous comment, this function is called, whereas the mock wasn't called. This should raise a red flag when writing tests.
```suggestion
        mock_update_user_overall_load.assert_called_once()
```

annotation/tests/test_tasks_services.py
BogicevicMilan: Ok.

annotation/tests/test_tasks_services.py
djo753: Look into `assert_called_once_with()` . You are going to need to save the user_id at the very least.

annotation/tests/test_tasks_services.py
```diff
@@ -0,0 +1,343 @@
...
+        "annotation.jobs.services.update_user_overall_load"
+    ) as mock_update_user_overall_load:
+        db_session = mock_session()
+        db_session.commit.side_effect = Exception("Commit failed")
+
+        with pytest.raises(Exception):
+            create_annotation_task(db_session, None)
+
+        db_session.commit.assert_not_called()
+        mock_update_user_overall_load.assert_not_called()
```
djo753, line 343: In this test we actually aren't testing an expected exception of this function. We are mocking the side_effect with an exception that the function doesn't declare in its code.

PR #907, opened 2024-08-15
test(annotation): cover jobs.services.py functions

Negoje:
cover for functions get_jobs_by_files and get_job_attributes_for_post


annotation/tests/test_jobs.py
```diff
@@ -88,3 +184,196 @@ def test_check_annotators(
...
+
+
+def test_collect_job_names_all_db():
+    mock_session = MagicMock()
+    mock_job = namedtuple("mock_job", ["job_id", "name"])
+    mock_session.query().filter().all.return_value = [
+        mock_job(1, "test1"),
+        mock_job(2, "test2"),
+    ]
+    expected_result = {1: "test1", 2: "test2"}
```
UkiStefanovic, line 196: Maybe make this test setup a fixture, since it appears in this and in the next test

UkiStefanovic (changes requested):
Everything else seems okay

annotation/tests/test_jobs.py
Negoje: You are right there was no need to use in both tests mock_job

annotation/tests/test_jobs.py
```diff
@@ -182,12 +186,11 @@ def test_check_annotators(
         check_annotators(annotators, validation_type)
 
 
-def test_collect_job_names_all_db():
+def test_collect_job_names_all_db(jobs_to_test_progress: Tuple[Job]):
```
cakeinsauce, line 189:
```suggestion
def test_collect_job_names_all_db(jobs_to_test_progress: Tuple[Job, ...]):
```
[link](https://docs.python.org/3/library/typing.html#:~:text=To%20denote%20a%20tuple%20which%20could%20be%20of%20any%20length%2C%20and%20in%20which%20all%20elements%20are%20of%20the%20same%20type%20T%2C%20use%20tuple%5BT%2C%20...%5D.%20To%20denote%20an%20empty%20tuple%2C%20use%20tuple%5B()%5D.%20Using%20plain%20tuple%20as%20an%20annotation%20is%20equivalent%20to%20using%20tuple%5BAny%2C%20...%5D%3A)

annotation/tests/test_jobs.py
```diff
@@ -276,3 +279,100 @@ def test_get_tasks_to_delete(tasks: Tuple[ManualAnnotationTask]):
...
+            2,
+            (Job.name, Job.status),
+            ("job_name_2", JobStatusEnumSchema.finished),
+        ),
+        (3, (Job.name,), None),
+    ),
+)
+def test_get_job_attributes_for_post(
+    job_id: int,
+    attributes: Union[Tuple[Base], Tuple[InstrumentedAttribute, ...]],
```
cakeinsauce, line 344: Could you clarify Tuple[Base]? I cannot really see that anywhere

annotation/tests/test_jobs.py
```diff
@@ -276,3 +279,100 @@ def test_get_tasks_to_delete(tasks: Tuple[ManualAnnotationTask]):
...
+    elif job_id == 2:
+        mock_session.query().filter_by().first.return_value = (
+            "job_name_2",
+            JobStatusEnumSchema.finished,
+        )
+        result = get_job_attributes_for_post(
+            mock_session, job_id, "test", attributes
+        )
+        assert result == expected_result
+    elif job_id == 3:
```
cakeinsauce, line 373: remove ifs; split into several tests

annotation/tests/test_jobs.py
Negoje: It means the Base class for Job model, but now that the code is refactored there is no need for that

cakeinsauce (approved):
great!

PR #912, opened 2024-08-20
test(annotation): cover check null fields

djo753:


annotation/tests/test_annotations.py
```diff
@@ -482,3 +591,582 @@ def test_construct_annotated_doc_db_error(
...
+    ), patch(
+        "annotation.annotations.main.logger_.exception"
+    ) as mock_logger, pytest.raises(
+        Exception
+    ):
+        construct_particular_rev_response(annotated_doc)
+        mock_logger.assert_called_once_with("Can't load annotation")
+
+
+def test_check_null_fields(annotation_doc_for_save: DocForSaveSchema):
```
BogicevicMilan, line 1164: Looks good.

PR #919, opened 2024-08-23
test(annotation): cover create_tasks

UkiStefanovic:
This pr contains tests for testing the function _create_tasks_ in the file /tests/test_distribution.py. 

The test covers two cases of the functions use. First case is when the amount of pages to assign is less than the MAX_PAGES attribute and the second case covers when the files have to be split to multiple tasks.


Negoje (approved):
Looks good

cakeinsauce (approved):
LGTM!

PR #920, opened 2024-08-23
test(annotation): cover module distribution

UkiStefanovic:
This pr adds new tests to cover the module annotation/distribution/main.py

All unit tests for the function in this module can be found in /tests/test_distribution.py


Negoje (approved):
Seems good to me

annotation/tests/test_distribution.py
```diff
@@ -1905,3 +1924,25 @@ def test_redistribute_error(
...
+        return_value=validation_files_pages,
+    ), patch("annotation.distribution.main.SPLIT_MULTIPAGE_DOC", True):
+        create_partial_validation_tasks(
+            {},
+            [{"user_id": "user1"}, {"user_id": "user2"}],
+            {},
+            1,
+            validation_tasks,
+            TaskStatusEnumSchema.pending,
+            None,
```
djo753, line 1943: This is a bool value, we should indicate this within the test, even if it is not utilized anywhere else except the mocked function. Pass in either True or False

annotation/tests/test_distribution.py
```diff
@@ -1905,3 +1924,25 @@ def test_redistribute_error(
...
+        )
+
+        assert len(validation_tasks) == 2
+        assert validation_tasks[0]["file_id"] == 1
+        assert len(validation_tasks[0]["pages"]) == 50
```
djo753, line 1948: Since we are using `SPLIT_MULTIPAGE_DOC` flag, we should assert both of the validation tasks and convey the splitting.
`assert len(validation_tasks[1]["pages"]) == 30` or even `assert validation_tasks[1]["pages"] == list(range(51,81))`

cakeinsauce (approved):
LGTM!

PR #932, opened 2024-08-24
test(annotation): cover common objs functions

BogicevicMilan:


annotation/tests/test_tasks_services.py
```diff
@@ -1632,3 +1632,50 @@ def test_remove_ids():
     assert "id" not in result
     assert "links" not in result
     assert "children" not in result
+
+
+def test_find_common_objs(setup_data: Dict[int, List[Tuple[Dict[str, str]]]]):
```
cakeinsauce, line 1637: simplify dict value typing to any

Issue #936, opened 2024-08-26
Remove lru-dict dependency from annotation service


cakeinsauce:
@djo753
Annotation service uses lru-dict = "1.2.0" dependency which implements dict like structure with LRU cache functionallity. We should remove it as this dependency not well maintained and from unknown developer.

A/C:

- [ ] lru-dict dependency is removed from poetry dependencies list
- [ ] annotation/annotation/tasks/services.py LRU dict cache usage is replaced with new implementation using built in python capabilities

DOD:
- [ ] lru-dict is removed
- [ ] code is covered with tests
- [ ] functionallity is tested manually with fully working badgerdoc

djo753:
Will do!

djo753:
Published a PR with the changes needed to cover this issue:
https://github.com/epam/badgerdoc/pull/940

PR #940, opened 2024-08-27
fix(tasks): add LRU class, remove lru-dict dependency

djo753:
PR regarding the issue: https://github.com/epam/badgerdoc/issues/936


annotation/annotation/tasks/services.py
```diff
@@ -62,6 +61,37 @@
...
+    def __getitem__(self, key):
+        if key in self.cache:
+            self.lru[key] = self.counter
+            self.counter += 1
+            return self.cache[key]
+        raise KeyError
+
+    def __setitem__(self, key, value):
+        if len(self.cache) >= self.capacity:
+            old_key = min(self.lru.keys(), key=lambda k: self.lru[k])
```
BogicevicMilan, line 87: The implementation isn't efficient enough. Try to avoid using the min function and aim for insert and delete operations to be in O(1).

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +61,37 @@
...
+
+class LRU:
+    def __init__(self, capacity):
+        self.capacity = capacity
+        self.counter = 0
+        self.cache = {}
+        self.lru = {}
+
+    def __contains__(self, item):
+        if item in self.cache:
```
BogicevicMilan, line 73: The counter should be updated even when checking if an item is in the cache.

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +61,37 @@
...
+            return True
+        else:
+            return False
+
+    def __getitem__(self, key):
+        if key in self.cache:
+            self.lru[key] = self.counter
+            self.counter += 1
```
BogicevicMilan, line 81: The counter can be incremented only up to the capacity; you don't need to go beyond that. Use modulo for this.

BogicevicMilan: Smth like priority queue

annotation/annotation/tasks/services.py
djo753: Due to the suggestion, I will rewrite the implementation using `OrderedDict` since it has O(1) methods for insert and delete, as it is a double linked list.

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,33 @@
     ValidationSchema,
 )
 
+
+class LRU(OrderedDict):
+    def __init__(self, capacity):
+        self.capacity = capacity
```
cakeinsauce, line 68: You need to invoke OrderedDict init as well

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,33 @@
...
+        if super().__contains__(item):
+            self.move_to_end(item)
+            return True
+        else:
+            return False
+
+    def __getitem__(self, key):
+        if key in self:
+            self.move_to_end(key)
+            return super().__getitem__(key)
```
cakeinsauce, line 80:
```suggestion
            return self[key]
```

cakeinsauce: but better solution would be first get the item first (if it is not found - exception is raised anyway, then move to end and return the found item. More pythonic approach. Otherwise you do extra work and raise exception without any details

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,33 @@
...
+        raise KeyError
+
+    def __setitem__(self, key, value):
+        if key in self:
+            self.move_to_end(key)
+        super().__setitem__(key, value)
```
cakeinsauce, line 86:
```suggestion
        self[key] = value
```

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,33 @@
...
+        if len(self) > self.capacity:
+            oldest = next(iter(self))
+            del self[oldest]
```
cakeinsauce, line 90:
```suggestion
            self.popitem(last=False)
```
because you move everything new to the right, meaning oldest is on the left

annotation/annotation/tasks/services.py
djo753: Cannot do that, it creates infinite recursion, since we are calling the method we are overriding.

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,34 @@
...
+    def __init__(self, capacity):
+        super().__init__
+        self.capacity = capacity
+
+    def __contains__(self, item):
+        if super().__contains__(item):
+            self.move_to_end(item)
+            return True
+        else:
+            return False
```
cakeinsauce, line 76: I dont think we need it. in operator just checks if item exists in the container, but it is not a retreival, so we shouldn't move it to end, meaning we dont need to redefine the method

annotation/tests/test_tasks_services.py
```diff
@@ -1472,3 +1472,12 @@ def test_get_common_ids(
...
     assert result == expected_result
+
+
+def test_lru():
+    cache = services.LRU(2)
+    cache[1] = "test1"
+    cache[2] = "test2"
+    cache[3] = "test3"
+    with pytest.raises(KeyError):
+        cache[1]
```
cakeinsauce, line 1483: split good and bad behavior into separate tests, please

annotation/annotation/tasks/services.py
djo753: It was done because of the idea of cache, where any interaction with an element in cache memory would be flagged as "used". I can remove it certainly.

annotation/annotation/tasks/services.py
djo753: I am getting a KeyError whenever I try to use the suggested implementation. It seems like that popitem calls the __getitem__ method after it has already removed the item from the dictionary. An error occurs in the __getitem__ method, after checking if the key is in the dictionary. 
```
    def test_lru():
        cache = services.LRU(2)
        cache[1] = "test1"
        cache[2] = "test2"
>       cache[3] = "test3"

tests\test_tasks_services.py:1481:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
annotation\tasks\services.py:84: in __setitem__
    super().popitem(last=False)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = LRU([(2, 'test2'), (3, 'test3')]), key = 1

    def __getitem__(self, key):
        if key in self:
>           self.move_to_end(key)
E           KeyError: 1

annotation\tasks\services.py:73: KeyError
```

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,26 @@
     ValidationSchema,
 )
 
+
+class LRU(OrderedDict):
+    def __init__(self, capacity):
```
cakeinsauce, line 67: add typing, please

annotation/annotation/tasks/services.py
cakeinsauce: Got it, super()[key] = something would work, but not that clear

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,26 @@
...
+
+    def __getitem__(self, key):
+        value = super().__getitem__(key)
+        self.move_to_end(key)
+        return value
+
+    def __setitem__(self, key, value):
+        if key in self:
+            self.move_to_end(key)
+        super().__setitem__(key, value)
```
cakeinsauce, line 79: what if we reorder them? first - set and then move to end. this way we dont have to check if key exists

annotation/annotation/tasks/services.py
```diff
@@ -62,6 +62,26 @@
     ValidationSchema,
 )
 
+
+class LRU(OrderedDict):
+    def __init__(self, capacity):
+        super().__init__()
+        self.capacity = capacity
```
cakeinsauce, line 69: 1. initialize child class
2. call superclass init

annotation/annotation/tasks/services.py
cakeinsauce: I guess it happens because we reimplement dunder methods instead of using normal names (like get, set). A lot of side effects

annotation/tests/test_tasks_services.py
```diff
@@ -1472,3 +1472,20 @@ def test_get_common_ids(
...
+    cache[2] = "test2"
+    assert cache[1] == "test1"
+    assert cache[2] == "test2"
+
+
+def test_lru_key_removed_due_to_capacity():
+    cache = services.LRU(2)
+    cache[1] = "test1"
+    cache[2] = "test2"
+    cache[3] = "test3"
```
cakeinsauce, line 1489: add validation that len after third inserted element is still two

cakeinsauce (approved):
Great work, test it locally in badgerdoc, please

PR #946, opened 2024-08-28
feat: add ci to badgerdoc_storage

khyurri:


lib/tenants/src/dependency.py
```diff
@@ -29,64 +26,14 @@ def __init__(
         Args:
             key: a private key for decoding tokens with hs256 alg
             algorithm: an alg for tokens, will be checked in available algorithms  # noqa
-            url: an url to auth service (http://bagerdoc-keycloack, http://dev1.gcov.ru)  # noqa
+            url: an url to auth service  # noqa
```
cakeinsauce, line 29:
```suggestion
            url: a url to auth service  # noqa
```

users/pyproject.toml
```diff
@@ -19,10 +19,10 @@ email-validator = "^1.1.3"
 apscheduler = "^3.9.1"
 pyjwt = "^2.6.0"
 badgerdoc-storage = { path = "../lib/badgerdoc_storage" }
+black = {path = "../lib/tenants"}
```
cakeinsauce, line 22: Could you elaborate on this? It should be tenants, not black, right?

cakeinsauce (changes requested):

.env.example
```diff
@@ -20,21 +20,19 @@ POSTGRES_DB=badgerdoc
 STORAGE_PROVIDER=minio
 
 # Minio configuration
-# In case of public host differs from minio internal address
-# setup this value
-MINIO_PUBLIC_HOST=
+# This MINIO_PUBLIC_HOST must be accessible from your browser
+MINIO_PUBLIC_HOST=http://badgerdoc-minio:9000
```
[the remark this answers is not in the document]
khyurri: Port is correct here, but it has to be exposed in `docker-compose-dev.yaml`
```
      - 9000:9000
      - 9001:9001
```

.github/workflows/badgerdoc-storage.yml
```diff
@@ -0,0 +1,32 @@
+name: badgerdoc_storage
```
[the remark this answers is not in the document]
khyurri: I want to rename all CI GH Actions to match the microservice/library name to avoid job names like the ones in the example:
Screenshot 2024-09-02 at 22 29 38

Example: 
`badgerdoc_storage`, `users`, `assets`, etc. It's much easier to understand which CI is failed.

users/pyproject.toml
khyurri: Yeah, it's weird dependency, removed at all. Thank you for catching it.

PR #955, opened 2024-09-02
feat: Jobs add datasets

ervandagadzhanyan:


assets/assets/schemas.py
```diff
@@ -55,18 +55,18 @@ class FileResponse(BaseModel):
     last_modified: datetime.datetime
     status: FileProcessingStatus
     path: str
-    datasets: List[str]
+    datasets: List[dict]
```
khyurri, line 58: Let's specify `datasets` 
```
class Dataset(BaseModel):
    id: int
    name: str
```

khyurri (changes requested):

assets/assets/schemas.py
ervandagadzhanyan: Thank you. I fixed.

PR #960, opened 2024-09-04
fix: fix KeyError

ervandagadzhanyan:


jobs/jobs/utils.py
```diff
@@ -291,21 +291,26 @@ def files_data_to_pipeline_arg(
...
+            pipeline_file: pipeline.PipelineFile = {
+                "bucket": file["bucket"],
+                "input": pipeline.PipelineFileInput(job_id=job_id),
+                "input_path": file["file"],
+                "pages": file["pages"],
+                "file_id": file_id,
+                "datasets": file["datasets"],
+            }
+        except KeyError as err:
+            logger.error(err)
```
khyurri, line 307: The use of `error` will not provide sufficient information about the exception here. It would be better to use `logger.exception('Unable to process file')`; however, if you use `continue` in this context, some files may be skipped in the pipeline event without notifying the user.

You need to choose a different strategy:
1. You can include a message in the success response indicating that some files were not processed.
2. You can return an error for the entire request.

khyurri (approved):

PR #985, opened 2024-10-01
fix: delete files

ervandagadzhanyan:


assets/assets/routers/files_router.py
```diff
@@ -7,6 +7,7 @@
 import minio
 import sqlalchemy.orm
 import sqlalchemy_filters.exceptions
+from badgerdoc_storage.storage import get_storage
```
cakeinsauce, line 10: import the storage instead of a dedicated function, please

assets/assets/utils/minio_utils.py
```diff
@@ -236,15 +236,19 @@ def upload_thumbnail(
...
+        if not objects:
             logger_.error(f"{obj} does not exist in bucket {bucket}")
             return False
-        for name in names:
-            client.remove_object(bucket, name)
+        for name in objects:
+            storage.remove(name)
+    except NotImplementedError as e:
+        logger_.error(f"{storage.__class__}.remove: {e}")
+        return False
```
cakeinsauce, line 251: I dont think we need it here

assets/assets/utils/minio_utils.py
```diff
@@ -236,15 +236,19 @@ def upload_thumbnail(
     return True
 
 
-def delete_one_from_minio(bucket: str, obj: str, client: minio.Minio) -> bool:
+def delete_one_from_minio(
```
cakeinsauce, line 239: could you rename it to something more generic, considering the fact we are not linked to minio

ervandagadzhanyan: Done

assets/assets/utils/minio_utils.py
ervandagadzhanyan: Done

cakeinsauce (approved):
Great!

PR #990, opened 2025-02-11
fix: S3 issues and jobs psql credentials

templier2:
1) Fix: creating resource configuration
When an Access Key is specified in .env, Badgerdoc sets the configuration for the Boto3 resource (region, endpoint, etc.). When no Access Key is provided (using an Instance Profile), the default region us-east-1 is used, and environment variables (region/endpoint) are ignored.

2) Fix: using signature v4 for S3 resource
When you provide an Access Key/Secret Key, SigV4 is used to generate pre-signed URLs for any region. With an Instance Profile, it relies on the configuration of the Boto3 resource and defaults to SigV2. SigV2 is accepted for old regions like the default region us-east-1, but not for most newer regions.

3) Fix: creating new bucket
Creating an S3 bucket via Boto3 requires the LocationConstraint parameter with the region specified. If you try to create a bucket in any region other than us-east-1 without this parameter, you'll encounter an error (regardless of whether the bucket already exists).

4) Fix: wrong jobs PSQL credentials
Unfortunately, jobs uses wrong creds for connecting to the DB.

5) Add some documentation about connecting S3.


templier2:
wil be recreated later

khyurri (approved):

PR #991, opened 2025-02-24
feat: add web service to host ml models for NER

apaplaus:
- Add small web service to Badgerdoc services that hosts ML models for NER (currently only Gliner models supported)
- Add new service to nginx config
- Add new service to docker-compose file
- Add separate build step into Makefile for the service


khyurri (approved):

.gitattributes
cakeinsauce, line 1: Do you need that? Usually IDEs have this setting in the right bottom corner. 
If needed, add a new line at the end, please

ml_server/README.md
```diff
@@ -0,0 +1,32 @@
...
+Workflow of the server is simple:
+ - You send POST request with json containing 4 mentioned parameters
+ - Server checks if model with given ID already downloaded and initialized. If yes, it's retrieved from cache. If not, model is downloaded, initialized and cached. Which may take some time, therefore if it's first run with this model, request might time out.
+ - Server runs model with other parameters from the request.
+ - Wraps result generation (usually it's a list of dictionaries with found entities) into dict with "entities" key
+ - Returns response to the client
+
+## How to run
+
+1. Make sure you have python 3 and pip installed
```
cakeinsauce, line 21: python >3.12

ml_server/requirements.txt
cakeinsauce, line 1: We use poetry for services and setuptools for libraries. It appears to be a service, so let's convert the builder to poetry, add pyproject.toml and have an ability to install service as a dependency (poetry install).
Additionally, because it is a service we should specify strict versions or minor compatability (== and ~= (^= in poetry as I recall))

cakeinsauce: Could we add dev deps and and quality gate tools. Feel free to grab pre commit from annotation service. And CI for github actions

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+from pathlib import Path
+from fastapi import FastAPI, HTTPException
+from pydantic import BaseModel
+from typing import List
+from functools import lru_cache
+
+from fastapi import Depends
+from gliner import GLiNER
+
+from tenant_dependency import get_tenant_info
```
cakeinsauce, line 11: You did not specify this dependency in project requirements

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+
+KEYCLOAK_HOST = os.getenv("KEYCLOAK_HOST", "")
+TENANT = get_tenant_info(url=KEYCLOAK_HOST, algorithm="RS256")
+ROOT_PATH = os.environ.get("ROOT_PATH", default="")
```
cakeinsauce, line 15: Worth adding .env.example file for this service
And also to example in the root of the project

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+# Cache will save up to this number of most recently used models. Larger amount will consume more memory,
+# but allows more models to be initialized at the same time.
+NUM_MODELS_TO_CACHE = 3
```
cakeinsauce, line 20: Not sure if we need this comment, var name self explanatory.
Maybe name it MODELS_CACHE_SIZE?

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+
+def get_version() -> str:
+    default = "0.1.0"
+    ver = Path(__file__).parent.parent / "version.txt"
```
cakeinsauce, line 24: Cannot see this file here.

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+    if ver.exists() and ver.is_file():
+        with open(ver, "r", encoding="utf-8") as file:
+            line = file.readline().strip()
+            return line or default
```
cakeinsauce, line 28: Let's use try except instead of exists ....

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+    title="ML models hosting server",
+    version=get_version(),
+    root_path=ROOT_PATH,
+    dependencies=[Depends(TENANT)])
+
+# Request body model
+class PredictionRequest(BaseModel):
+    model_id: str
+    threshold: float
+    labels: List[str]
```
cakeinsauce, line 42: let's use modern style
```suggestion
    labels: list[str]
```

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+app = FastAPI(
+    title="ML models hosting server",
+    version=get_version(),
+    root_path=ROOT_PATH,
+    dependencies=[Depends(TENANT)])
+
+# Request body model
+class PredictionRequest(BaseModel):
+    model_id: str
+    threshold: float
```
cakeinsauce, line 41: can it be negative? if no, let's add this requirement

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+
+    Returns:
+        GLiNER: initialized Gliner model
+    """
+    try:
+        # Initialize and cache the model
+        model = GLiNER.from_pretrained(model_id)
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=f"Error loading model '{model_id}': {str(e)}")
+
```
cakeinsauce, line 61:
```suggestion
```

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+    Args:
+        model_id (str): ID of the model to download
+
+    Returns:
+        GLiNER: initialized Gliner model
+    """
+    try:
+        # Initialize and cache the model
+        model = GLiNER.from_pretrained(model_id)
+    except Exception as e:
```
cakeinsauce, line 59: Can we except more specific exception? This one is too general. Additionally would be nice to have more explanatory http exception. Something like "Could not load model model_id=...".
The one that comes to mind when you haven't found the model. And it should be 404 in this case.
Don't expose exception to outside. Rather raise http exception from caught one using dedicated syntax

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+        raise HTTPException(status_code=500, detail=f"Error loading model '{model_id}': {str(e)}")
+    
+    return model
+
+
+# Named Entity Recognition endpoint
```
cakeinsauce, line 65: Let's name the router function that way and remove the comment

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+@app.post("/ner")
+async def ner(request: PredictionRequest):
+    model = get_model(request.model_id)
+
+    # Perform prediction
```
cakeinsauce, line 70:
```suggestion
```

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+    try:
+        entities = model.predict_entities(request.text, request.labels, threshold=request.threshold)
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=f"Error during prediction: {str(e)}")
```
cakeinsauce, line 74: Same as previous comment

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Return the prediction results
```
cakeinsauce, line 76:
```suggestion
```

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+    return {"entities": entities}
+
+# Run the application with: uvicorn main:app --reload
```
cakeinsauce, line 79:
```suggestion
```

cakeinsauce (changes requested):
Please address to comments above

ml_server/main.py
```diff
@@ -0,0 +1,79 @@
...
+    model = get_model(request.model_id)
+
+    # Perform prediction
+    try:
+        entities = model.predict_entities(request.text, request.labels, threshold=request.threshold)
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=f"Error during prediction: {str(e)}")
+
+    # Return the prediction results
+    return {"entities": entities}
```
cakeinsauce, line 77: Let's create a Pydantic schema for response for swagger documentation

.gitattributes
apaplaus: It's more of setting for git to properly adjust file endlines. Because without it I had issues for several times when line ending in bash scripts were replaced with CRLF after git sync. And then image got broken after rebuild.

ml_server/main.py
apaplaus: Then this file will contain only one entry, with KEYCLOAK_HOST. 
And it's already in .env.example in the root of the project. 
And ROOT_PATH is always service-specific. So you already have it set-up in compose file, and if you don't use compose and want to just run this service, it will still work, because you won't have proxy.

ml_server/main.py
apaplaus: This comment is more for explaining why this var exists and what will happen if someone want to change it, just in case.

ml_server/main.py
apaplaus: I copied this function from other services, where it's always the same.

ml_server/main.py
apaplaus: I didn't want to spend time going and searching the source code of the library for all possible exceptions that Gliner can raise. That's why I used general exception, basically you still will get info about original exception in the output, and for us as a user of that library the most important part is that it's failed and we can't proceed.

ml_server/main.py
apaplaus: I've added it to Readme. And as I understand other services do the same. Because you either use python_base image (that already have those installed) or you install it locally then you go by Readme.

ml_server/main.py
apaplaus: added

.github/workflows/ml-server.yml
```diff
@@ -0,0 +1,25 @@
...
+      matrix:
+        python-version: ["3.12.1"]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "pip"
+      - run: pip install flake8
```
cakeinsauce, line 24: lets install using poetry dev dependencies. feel free to address to other services actions file

ml_server/.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+    hooks:
+      - id: black
+        args:
+          - --line-length=79
+  - repo: https://github.com/pycqa/flake8
+    rev: 7.1.2
+    hooks:
+      - id: flake8
+        args:
+          - --extend-ignore=E203
```
cakeinsauce, line 22: let's add these checks to CI as well

cakeinsauce: Additionally, no need to specify args for linters here, they should have been parsed from pyproject.toml

ml_server/Dockerfile
```diff
@@ -0,0 +1,23 @@
+ARG base_image=818863528939.dkr.ecr.eu-central-1.amazonaws.com/badgerdoc/python_base_3.12:0.2.0
+FROM ${base_image} as build
+
+# Without it pip show warnings when installing packages as root
+ENV PIP_ROOT_USER_ACTION ignore
```
cakeinsauce, line 5: We shouldn't really care about this, from my point of view useless here. Up to you

ml_server/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+    && poetry install --no-root --only main \
+    && :
+
+CMD uvicorn main:app --port 8080 --host 0.0.0.0
+
+EXPOSE 8080
+
+FROM build as development
+
+RUN poetry install --no-root
```
cakeinsauce, line 23: add a new line at the end, please

ml_server/main.py
cakeinsauce: Could you return GlinerResponse object here to make it explicit, please?

ml_server/pyproject.toml
```diff
@@ -0,0 +1,44 @@
...
+
+
+[tool.black]
+line_length = "79"
+
+[tool.isort]
+profile = "black"
+line_length = "79"
+
+[tool.pylint.settings]
```
cakeinsauce, line 29: we dont have pylint here

cakeinsauce (changes requested):
ptal at the comments below

.github/workflows/ml-server.yml
apaplaus: Every workflow is using this, and I don't want to change the way how it's done for one service.

ml_server/.pre-commit-config.yaml
apaplaus: I used CI config from other services, Unfortunately I don't have time for setting up and testing new CI pipelines for a service and syncing it with pyproject.toml. All other services are having this logic, I took it from there.

Issue #993, opened 2025-03-10
When creating a job from revision files are not shown in the job info


cakeinsauce:
When creating a job from a file revision files are not displayed in the job info
![Image](https://github.com/user-attachments/assets/a9a5e9ca-4f2b-44df-ad24-c4830f1341be)

After starting and finishing the job it stays the same
![Image](https://github.com/user-attachments/assets/d9ee0acb-0f08-4024-bd17-58a6653ec955)

How to reproduce:
1. Create a manual job (for simplicity) and start it
2. Finish the job (add sample annotation for visibility in ui)
3. Open the job from Jobs tab and select the file from Files tab 
4. On the right select Information tab and the only revision (it was created on the step 2) and press New Job from Revision
5. Here just create a manual job.
6. Observe that list of files is not loading up

halilaydin242188:
i will look at this issue

halilaydin242188:
![Image](https://github.com/user-attachments/assets/e36f8a72-b587-4692-81ce-f6a4137ae5aa)

when a job is created from a revision, web does not send a file id to the backend, if the file id is sent, the file can be seen normally

PR #996, opened 2025-03-12
fix: files are not showing up in job review page when job is created from a revision

halilaydin242188:
* fix files are not showing up in job review page when job is created from a revision
- the 'file_id' is missing when job created from a revision, now it's sent

closes: #993


web/src/connectors/edit-job-connector/edit-job-connector.tsx
```diff
@@ -246,7 +247,7 @@ const EditJobConnector: FC<EditJobConnectorProps> = ({
                 name: jobName,
                 revisions: revisionId ? [revisionId] : revisions,
                 datasets,
-                files,
+                files: (files?.length ?? 0) > 0 ? files : fileId ? [fileId] : [],
```
cakeinsauce, line 250: what about this? will it work?
```suggestion
                files: files?.length ? files : fileId ? [fileId] : [],
```

cakeinsauce (approved):
LGTM! Please remove gh issue number from PR title and add it to PR description

Issue #997, opened 2025-03-13
Creating new job has some lacking validations such as file_id and job_name


halilaydin242188:
- When creating a new job, even though the file with file_id does not exist, it allows it to create the job. This works for the job name too.

Recreation steps: 

- prepare a create_job request in postman (to simplify i copied a request from web ui and pasted it into postman)
- in post body, give files field a non existing file_id
- it creates the job and return 200 even though file does not exist.

![Image](https://github.com/user-attachments/assets/33fcfcbc-cfe1-48ca-be80-4fcfded01474)

halilaydin242188:
i'll look into this

PR #998, opened 2025-03-14
fix: create_job has no check for dublicate job name and files exist or not 

halilaydin242188:
add validations for:
- check if the given referred files exist in database
- check if the job name is already in use

closes: #997

### DIFF
--- a/jobs/jobs/db_service.py
+++ b/jobs/jobs/db_service.py
@@ -226,11 +226,11 @@ def change_job(
     db.commit()
 
 
-def is_job_exists_by_name(db: Session, job_name: str) -> bool:
-    """Check if a job record exist in db with the 'job_name'"""
-    job_count_by_name = (
+def get_jobs_by_name(db: Session, job_name: str) -> List[dbm.CombinedJob]:
+    """Retrieve jobs from database by name"""
+    result = (
         db.query(dbm.CombinedJob)
         .filter(dbm.CombinedJob.name == job_name)
-        .count()
+        .all()
     )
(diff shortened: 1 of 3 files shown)


jobs/jobs/db_service.py
```diff
@@ -224,3 +224,13 @@ def change_job(
         if new_job_params_dict[job_param_key] is not None:
             setattr(job, job_param_key, new_job_params_dict[job_param_key])
     db.commit()
+
+
+def is_job_exists_by_name(db: Session, job_name: str) -> bool:
```
cakeinsauce, line 229: let's make this function more generic - get job by name. You could utilize that in numerous places later. No need to check for existence here

jobs/jobs/main.py
```diff
@@ -60,6 +60,32 @@ async def create_job(
...
+    if len(job_params.files) > 0:
+        _, matched_file_ids_in_db = (
+            await utils.get_files_data_from_separate_files(
+                job_params.files, current_tenant, token_data.token
+            )
+        )
+
+        non_existing_file_ids = list(
+            set(job_params.files).difference(matched_file_ids_in_db)
+        )
```
cakeinsauce, line 81: from a brief look it can be simplified to a function which takes ids of the files and fetch them from DB. If list of files returned by the function is less than the input - we are missing some of them

cakeinsauce (approved):

cakeinsauce (changes requested):
ptal

cakeinsauce (approved):

Issue #1003, opened 2025-03-18
Category related issues


halilaydin242188:
1.  categories can be set cyclically parent: 

![Image](https://github.com/user-attachments/assets/3325188b-dd46-4d52-9e13-3b94257ba5bd)

2. when you change the parent of a category that is used in job, job's task ui, categories can not be retrieved (annotation/jobs/86/categories/search) return 404

PR #1004, opened 2025-03-18
fix: category issues

halilaydin242188:
- check cyclic parent error while updating a category

closes: #1003

### DIFF
--- a/annotation/annotation/categories/resources.py
+++ b/annotation/annotation/categories/resources.py
@@ -160,8 +160,8 @@ def update_category(
     except ValueError as error:
         raise HTTPException(
             status_code=400,
-            detail=f"{error}",
-        )
+            detail="Selected parent is invalid due to being a child of the target category",
+        ) from error
 
     if not category_db:
         raise NoSuchCategoryError("Cannot update category parameters")
--- a/annotation/annotation/categories/resources.py
+++ b/annotation/annotation/categories/resources.py
@@ -153,9 +153,16 @@ def update_category(
     """
     Updates category by id and returns updated category.
     """
-    category_db = update_category_db(
-        db, category_id, query.dict(), x_current_tenant
-    )
+    try:
+        category_db = update_category_db(
+            db, category_id, query.dict(), x_current_tenant
+        )
+    except ValueError as error:
+        raise HTTPException(
+            status_code=400,
+            detail=f"{error}",
+        )
+
     if not category_db:
         raise NoSuchCategoryError("Cannot update category parameters")
     return response_object_from_db(category_db)
--- a/annotation/annotation/categories/resources.py
+++ b/annotation/annotation/categories/resources.py
@@ -160,7 +160,7 @@ def update_category(
     except ValueError as error:
         raise HTTPException(
             status_code=400,
-            detail="Selected parent is invalid due to being a child of the target category",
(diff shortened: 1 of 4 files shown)


annotation/annotation/categories/services.py
```diff
@@ -462,3 +466,15 @@ def combine_categories(
             if cat:
                 pages_categories.add(cat)
     return categories | pages_categories
+
+
+def check_cyclic_parenting(
+    db: Session, category_id: str, new_parent_category: Category
+):
```
cakeinsauce, line 473: Add the return typing, please

annotation/annotation/errors.py
```diff
@@ -47,6 +47,11 @@ def __init__(self, message: str):
         self.message = message
 
 
+class CyclicParentError(Exception):
+    def __init__(self, message: str):
+        self.message = message
```
cakeinsauce, line 52: I don't think we need a new exception for that. Please utilize built in exceptions, e.g. ValueError

annotation/annotation/errors.py
```diff
@@ -158,3 +163,10 @@ def agreement_score_service_error_handler(
         status_code=400,
         content={"detail": f"Agreement score service error. {exc}"},
     )
+
+
+def cyclic_parent_error_handler(request: Request, exc: CyclicParentError):
+    return JSONResponse(
+        status_code=400,
+        content={"detail": f"Cyclic parent error. {exc.message}"},
+    )
```
cakeinsauce, line 172: Are we sure it is a case for a lot of places? I think only when we create and modify, probably the handler is an overkill

cakeinsauce (changes requested):
ptal

annotation/annotation/errors.py
halilaydin242188: ValueError is not handled by fastapi right now so only 500 is returned to the ui. For this an exception handler for ValueError can be added. What do you think ?

image

annotation/annotation/categories/services.py
```diff
@@ -462,3 +466,15 @@ def combine_categories(
...
+    """check if the category is already the parent of the given parent_id"""
+    parents_of_parent_cat = fetch_category_parents(db, new_parent_category)
+    parent_ids_of_parent = [p.id for p in parents_of_parent_cat]
+    if category_id in parent_ids_of_parent:
+        raise CyclicParentError(
+            "This category is in the parent tree of the target parent"
+        )
```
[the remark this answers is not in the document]
halilaydin242188: i could not think of any other solution, if this is ok can we accept this code ?

annotation/annotation/categories/resources.py
```diff
@@ -153,9 +153,16 @@ def update_category(
...
-    )
+    try:
+        category_db = update_category_db(
+            db, category_id, query.dict(), x_current_tenant
+        )
+    except ValueError as error:
+        raise HTTPException(
+            status_code=400,
+            detail=f"{error}",
+        )
```
cakeinsauce, line 164: 1. Raise the exception from the original one (from syntax)
2. Use specific error message, do not expose internals to the end user

cakeinsauce (changes requested):
ptal

cakeinsauce (approved):
Great!

Issue #992, opened 2025-03-07
Progress bar doesn't work on airflow pipelines


cakeinsauce:
When I create a job with any Airflow pipeline selected, the progress bar in the right top corner of the job info doesn't progress despite that DAG in Airflow has finished its work. 
![Image](https://github.com/user-attachments/assets/0bb41934-007a-46ec-907d-0db173ac5596)

The endpoint responsible for job progression status throwing 422 error

![Image](https://github.com/user-attachments/assets/76ddba9f-f4d4-4d67-860d-48141f75f652)

![Image](https://github.com/user-attachments/assets/c3f66a6a-03ca-4340-91e4-faca52e7cf12)

How to reproduce:
1. Open Jobs tab
2. Select any document (upload if nothing there)
3. Click next, add job name and select any airflow pipeline
4. Click New Job
5. The new job will be created, progress bar won't ever finish (even if airflow task will) and in dev console you will see error with jobs endpoint progress bar.

AliBugday:
I'll look at this issue.

PR #1008, opened 2025-03-28
fix: progress bar endpoint for Extraction and ExtractionAndAnnotation job.

AliBugday:
closes: #992


jobs/jobs/utils.py
```diff
@@ -563,29 +566,81 @@ async def get_job_progress(
...
+            )
+    
     response.update({"mode": str(job.mode)})
-    return response  # type: ignore
+    
+    if job.mode == JobMode.Manual:
+        await update_manual_job(session, job, response)
+        return response
+
+    elif job.mode == JobMode.Automatic:
```
cakeinsauce, line 595: just if

jobs/jobs/utils.py
```diff
@@ -563,29 +566,81 @@ async def get_job_progress(
...
+        return await handle_automatic_job(session, job, response, job_id)
+    
+    
+    
+async def update_manual_job(session, job, response):
```
cakeinsauce, line 600: add typings, please

jobs/jobs/utils.py
```diff
@@ -563,29 +566,81 @@ async def get_job_progress(
...
+    """Updates the job status based on the DAG execution response."""
+    if response.get("finished", 0) > 0:
+        if response["finished"] == response["total"]:
+            db_service.update_job_status(session, job, Status.finished)
+        else:
+            db_service.update_job_status(session, job, Status.in_progress)
```
cakeinsauce, line 606:
```suggestion
    if response.get("finished", 0) == response["total"]:
        db_service.update_job_status(session, job, Status.finished)
    else:
        db_service.update_job_status(session, job, Status.in_progress)
```

jobs/jobs/utils.py
```diff
@@ -563,29 +566,81 @@ async def get_job_progress(
...
+    
+
+
+async def handle_automatic_job(session, job, response, job_id):
```
cakeinsauce, line 610: add typings, please

cakeinsauce: This function is too complex, please split this up into smaller chunks

cakeinsauce (changes requested):
PTAL

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +51,112 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+                updated_dag = DAG(is_paused=False)
+                api_instance.patch_dag(dag_id, updated_dag)
+                
+    except ApiException as e:
+        return f"error fetching DAG state: {str(e)}"
+    except Exception as e:
+        return f"Unexpected error: {str(e)}"
+    
+    
+async def wait_for_dag_completion_async(dag_id: str, dag_run_id: str, poll_interval=2, timeout=600):
```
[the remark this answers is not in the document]
AliBugday: I am suggesting that this function is necessary because some Airflow DAGs take a long time to complete and this function ensures that the DAGs are completed. If a DAG doesn't finish within a specified timeframe (e.g., 5 minutes), it raises a timeout error. Without this function, get_dag_status will return an inaccurate result while the DAG is still running, leading to incorrect program behavior.

jobs/jobs/utils.py
AliBugday: I would like to stick with my current implementation. We need to check if any task has been completed. In case no task is completed, Annotation tasks are in pending state and I think it would be better to wait for users to start the job from the website. With the approach you suggested, all newly created annotation tasks would immediately move to the "in progress" state, even if none have actually been started, which I think isn't ideal.

jobs/jobs/airflow_utils.py
```diff
@@ -31,11 +37,15 @@ def get_configuration():
     )
 
 
-def get_api_instance():
-    configuration = get_configuration()
-
+def get_api_instance(configuration):
     with client.ApiClient(configuration) as api_client:
-        yield api_client
+        return api_client
```
cakeinsauce, line 42: make it context manager. And I think client closes once you return it

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
         return dag_api.get_dags()
 
 
+async def get_dag_status(
+    dag_id: str, job_id: int, api_client: client.ApiClient
+) -> str:
+    """Fetches the DAG run status for the given DAG ID."""
+
+    # api_client = airflow_connection()
```
cakeinsauce, line 64: remove if not needed

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+        api_response = api_instance.get_dag_run(dag_id, dag_run_id)
+
+        dag_state = api_response.state.value.lower()
+        if dag_state == AirflowPiplineStatus.success.value:
+            return AirflowPiplineStatus.success.value
+        elif dag_state == AirflowPiplineStatus.failed.value:
+            return AirflowPiplineStatus.failed.value
+        else:
+            # if other states like "queued", "running", etc.
+            raise RuntimeError(f"Unexpected DAG run state: {dag_state}")
```
cakeinsauce, line 79: No need to put everything in try except, you should only handle code which throws the errors you specified in the block

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+
+    except ApiException as e:
+        raise RuntimeError(f"Error fetching DAG run status: {e}") from e
+    except Exception as e:
```
cakeinsauce, line 83: Let's avoid handling generic exception, specific the exact ones

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+        if dag_state == AirflowPiplineStatus.success.value:
+            return AirflowPiplineStatus.success.value
+        elif dag_state == AirflowPiplineStatus.failed.value:
+            return AirflowPiplineStatus.failed.value
+        else:
+            # if other states like "queued", "running", etc.
+            raise RuntimeError(f"Unexpected DAG run state: {dag_state}")
+
+    except ApiException as e:
+        raise RuntimeError(f"Error fetching DAG run status: {e}") from e
```
cakeinsauce, line 82: no need to add caught exception to a new one, we have a traceback for that. Unless you want to put specific info there, then do that

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+    try:
+        api_instance = DAGApi(api_client)
+
+        dag = api_instance.get_dag(dag_id)
+        if dag.is_paused:
+            updated_dag = DAG(is_paused=False)
+            api_instance.patch_dag(dag_id, updated_dag)
+    except ApiException as e:
+        raise RuntimeError(f"Failed to fetch or update DAG state: {e}") from e
+    except Exception as e:
```
cakeinsauce, line 97: Let's avoid handling generic exception, specific the exact ones

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+
+
+async def activate_dag(dag_id: str, api_client: client.ApiClient) -> None:
+    try:
+        api_instance = DAGApi(api_client)
+
+        dag = api_instance.get_dag(dag_id)
+        if dag.is_paused:
+            updated_dag = DAG(is_paused=False)
+            api_instance.patch_dag(dag_id, updated_dag)
```
cakeinsauce, line 94: Make sure we have code here which can throw errors specified in the except block, remove redundant if any

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+    except ApiException as e:
+        raise RuntimeError(f"Failed to fetch or update DAG state: {e}") from e
```
cakeinsauce, line 96: no need to add caught exception to a new one, we have a traceback for that. Unless you want to put specific info there, then do that

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+    except Exception as e:
+        raise RuntimeError(f"Unexpected error: {e}") from e
+
+
+async def wait_for_dag_completion_async(
+    dag_id: str,
+    dag_run_id: str,
+    api_client: client.ApiClient,
+    poll_interval=2,
+    timeout=600,
```
cakeinsauce, line 106: typings

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+            if dag_state in ["success", "failed"]:
+                return dag_state
+
+            await asyncio.sleep(poll_interval)
+
+    except ApiException as e:
+        raise RuntimeError(f"error fetching DAG state: {e}") from e
+    except Exception as e:
+        raise RuntimeError(f"Unexpected error: {e}") from e
+    raise TimeoutError("Timeout reached")
```
cakeinsauce, line 126: rather put try except inside of while block, this way it is much cleaner and you can move out some of the code from the try except if it is not handled by the block

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+            api_response = api_instance.get_dag_run(dag_id, dag_run_id)
+            dag_state = api_response.state.value
+
+            if dag_state in ["success", "failed"]:
+                return dag_state
+
+            await asyncio.sleep(poll_interval)
+
+    except ApiException as e:
+        raise RuntimeError(f"error fetching DAG state: {e}") from e
```
cakeinsauce, line 123: no need to add caught exception to a new one, we have a traceback for that. Unless you want to put specific info there, then do that

jobs/jobs/airflow_utils.py
cakeinsauce: And ideally to have these values in a configuration params. Check them out, maybe it already exists.

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+    api_client: client.ApiClient,
+    poll_interval=2,
+    timeout=600,
+) -> Literal["success", "failed"]:
+    """waiting for DAG completion. If DAG does not finish within the given time, a timeout error is returned."""
+    try:
+        api_instance = DAGRunApi(api_client)
+
+        start_time = time.time()
+        while time.time() - start_time < timeout:
```
cakeinsauce, line 113: btw you can do that with for loop, a bit cleaner, but not that explicit. Up to you 
```python
    num_iterations = int(timeout // poll_interval)
    for _ in range(num_iterations):
        ...
        await asyncio.sleep(poll_interval)
```

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+        raise RuntimeError(f"Unexpected error: {e}") from e
+
+
+async def wait_for_dag_completion_async(
+    dag_id: str,
+    dag_run_id: str,
+    api_client: client.ApiClient,
+    poll_interval=2,
+    timeout=600,
+) -> Literal["success", "failed"]:
```
cakeinsauce, line 107: I see you added `AirflowPiplineStatus` but did not use it here, let's fix that

cakeinsauce: You can still use literal, but with enum values

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,76 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+    """waiting for DAG completion. If DAG does not finish within the given time, a timeout error is returned."""
+    try:
+        api_instance = DAGRunApi(api_client)
+
+        start_time = time.time()
+        while time.time() - start_time < timeout:
+            api_response = api_instance.get_dag_run(dag_id, dag_run_id)
+            dag_state = api_response.state.value
+
+            if dag_state in ["success", "failed"]:
```
cakeinsauce, line 117: Same comment regarding enum AirflowPiplineStatus

jobs/jobs/utils.py
```diff
@@ -585,7 +590,147 @@ async def get_job_progress(
...
+    if job.mode == JobMode.Manual:
+        await update_manual_job_status(session, job, response)
+        return response
+
+    elif job.mode == JobMode.Automatic:
+        return await handle_automatic_job(
+            session, job, response, job_id, api_client
+        )
+
+    logger.warning(
```
cakeinsauce, line 602: great approach!

jobs/jobs/utils.py
```diff
@@ -585,7 +590,147 @@ async def get_job_progress(
         )
 
     response.update({"mode": str(job.mode)})
-    return response  # type: ignore
+    if job.mode == JobMode.Manual:
+        await update_manual_job_status(session, job, response)
+        return response
+
+    elif job.mode == JobMode.Automatic:
```
cakeinsauce, line 597: Still not fixed. Please utilize if, no need for elif. If first if block is entered, it will exit the function with return statement

jobs/jobs/utils.py
cakeinsauce: Are you sure it will work like that? I just suggested a refactoring of if statements which should not affect the logic at all. If I'm mistaken, then let's keep the original

jobs/jobs/airflow_utils.py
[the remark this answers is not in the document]
cakeinsauce: But how would it look like from the user perspective? They would wait for the pipeline to finish? What if pipeline runs 30 min. HTTP timeout would occur then. As I recall Front End polls the status endpoint periodically, and, if job is in progress it should return that, not wait for n minutes to finish

jobs/jobs/utils.py
```diff
@@ -585,7 +590,147 @@ async def get_job_progress(
...
+    response: Dict[str, int],
+    job_id: int,
+    api_client: client.ApiClient,
+) -> Dict[str, int]:
+    """Handles automatic jobs with or without a pipeline."""
+    if response.get("total", 0) == 0:
+        return await handle_pipeline_driven_job(
+            session, job, response, job_id, api_client
+        )
+    else:
```
cakeinsauce, line 632: no need for else here

jobs/jobs/utils.py
```diff
@@ -585,7 +590,147 @@ async def get_job_progress(
...
+    response: Dict[str, int],
+    job_id: int,
+    api_client: client.ApiClient,
+) -> Dict[str, int]:
+    """Handles automatic jobs that rely on pipeline DAG execution."""
+    pipeline_id = job.pipeline_id
+    dag_run_id = f"{pipeline_id}:{job_id}"
+
+    await activate_pipline(pipeline_id, api_client)
+    db_service.update_job_status(session, job, Status.in_progress)
```
cakeinsauce, line 649: Wait, so we start the job IF we request the pipeline progress? It does not sound right. User could create a job, but did not start it on purpose. If FE requests the job status we must not start it, or modify it in any way. Just get the status and return it. Usually pending status is returned in such cases

jobs/jobs/utils.py
```diff
@@ -585,7 +590,147 @@ async def get_job_progress(
...
+    try:
+        pipeline_result = await airflow_utils.get_dag_status(
+            pipeline_id, job_id, api_client
+        )
+    except RuntimeError as e:
+        logger.exception(
+            "Runtime error occurred while getting DAG status result"
+        )
+        db_service.update_job_status(session, job, Status.failed)
+        raise fastapi.HTTPException(
```
cakeinsauce, line 687: Do not expose internals to the user. Consider HTTPExceptions as messages to the end user. In your case something like "Failed to fetch pipeline status" is enough. User doesn't even know what dag is, for them it is a job with pipelines sometimes

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,45 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+        api_instance = DAGApi(api_client)
+
+        dag = api_instance.get_dag(dag_id)
+        if dag.is_paused:
+            updated_dag = DAG(is_paused=False)
+            api_instance.patch_dag(dag_id, updated_dag)
+    except ApiException:
+        raise RuntimeError(f"Failed to fetch or update DAG state. DAG id: {dag_id}")
+    except AttributeError:
+        raise RuntimeError(f"Error while passing attributes. DAG id: {dag_id}")
```
cakeinsauce, line 70:
```suggestion
    except ApiException as err:
        raise RuntimeError(f"Failed to fetch or update DAG state. DAG id: {dag_id}") from err
    except AttributeError as err:
        raise RuntimeError(f"Error while passing attributes. DAG id: {dag_id}") from err
```

jobs/jobs/airflow_utils.py
```diff
@@ -46,6 +56,45 @@ async def get_dags() -> List[pipeline.AnyPipeline]:
...
+) -> AirflowPipelineStatus:
+    try:
+        api_instance = DAGRunApi(api_client)
+        api_response = api_instance.get_dag_run(dag_id, dag_run_id)
+        dag_state = api_response.state.value
+
+    except ApiException:
+        raise RuntimeError(f"Error fetching DAG run status. Job id: {dag_id}")
+    except (ValueError, TypeError, AttributeError):
+        raise RuntimeError(f"Error ApiClient incorrectly configured. DAG id: {dag_id}")
```
cakeinsauce, line 86:
```suggestion
    except ApiException as err:
        raise RuntimeError(f"Error fetching DAG run status. Job id: {dag_id}") from err
    except (ValueError, TypeError, AttributeError) as err:
        raise RuntimeError(f"Error ApiClient incorrectly configured. DAG id: {dag_id}") from err
```

jobs/jobs/utils.py
```diff
@@ -581,9 +584,126 @@ async def get_job_progress(
             status_code=fastapi.status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail=f"Failed request to the Annotation Manager: {err}",
         )
-
+    
     response.update({"mode": str(job.mode)})
-    return response  # type: ignore
+
+    if job.type in [JobType.AnnotationJob, JobType.ExtractionJob, JobType.ExtractionWithAnnotationJob]:
```
cakeinsauce, line 590: use tuple

jobs/jobs/utils.py
```diff
@@ -581,9 +584,126 @@ async def get_job_progress(
...
+async def update_job_status_by_type(
+    session: Session,
+    job: CombinedJob,
+    response: Dict[str, int],
+    job_id: int,
+    api_client: client.ApiClient,
+) -> None:
+    job_type = job.type
+
+    if job_type in [JobType.AnnotationJob, JobType.ExtractionWithAnnotationJob]:
```
cakeinsauce, line 610: use tuple

jobs/jobs/utils.py
```diff
@@ -581,9 +584,126 @@ async def get_job_progress(
...
+
+    if pipeline_status == AirflowPipelineStatus.success.value:
+        db_service.update_job_status(session, job, Status.finished)
+        response["finished"] = 1
+    elif pipeline_status == AirflowPipelineStatus.failed.value:
+        db_service.update_job_status(session, job, Status.failed)
+        response["finished"] = 0
+    elif pipeline_status == AirflowPipelineStatus.running.value:
+        db_service.update_job_status(session, job, Status.in_progress)
+        response["finished"] = 0
```
cakeinsauce, line 671: you can set specific var in if statements and call function update_job_status at the end after them (reduce duplication)

jobs/jobs/utils.py
```diff
@@ -581,9 +584,126 @@ async def get_job_progress(
...
+            pipeline_id,
+            dag_run_id,
+            api_client=api_client,
+        )
+    except RuntimeError as e:
+        logger.exception(
+            f"Runtime error occurred while fetching pipeline status: {e}"
+        )
+        raise fastapi.HTTPException(
+            status_code=500, detail=f"Failed to fetch pipeline status."
```
cakeinsauce, line 692:
```suggestion
            status_code=500, detail="Failed to fetch pipeline status."
```

jobs/jobs/utils.py
```diff
@@ -581,9 +584,126 @@ async def get_job_progress(
...
+        )
```
cakeinsauce, line 693:
```suggestion
        ) from e
```

jobs/jobs/utils.py
```diff
@@ -581,9 +584,126 @@ async def get_job_progress(
...
+async def activate_pipeline(
+    pipeline_id: str, api_client: client.ApiClient
+) -> None:
+    try:
+        await airflow_utils.activate_dag(pipeline_id, api_client)
+    except RuntimeError as e:
+        logger.exception(f"Failed to activate DAG: {e}")
+        raise fastapi.HTTPException(
+            status_code=500, detail=f"Failed to activate pipline"
+        )
```
cakeinsauce, line 706:
```suggestion
        ) from e
```

jobs/tests/fixtures_schemas.py
```diff
@@ -59,3 +59,23 @@ def mock_AnnotationJobParams2():
...
+        pipeline_engine="Airflow",
+        annotators=["annotator1", "annotator2"],
+        validators=["validator1", "validator2"],
+        owners=["owner1"],
+        categories=["category1", "category2"],
+        is_auto_distribution=False,
+        deadline=datetime.datetime.utcnow() + datetime.timedelta(days=1),
+        validation_type="cross",
+    )
+    return mockExtractionAnnotationJobParams
```
cakeinsauce, line 81: 1. no need for extra var, just return that
2. pytest recommends yielding from fixtures (https://docs.pytest.org/en/stable/how-to/fixtures.html#yield-fixtures-recommended)
3. add a new line at the end of the file

jobs/tests/fixtures_schemas.py
```diff
@@ -59,3 +59,23 @@ def mock_AnnotationJobParams2():
         deadline=datetime.datetime.utcnow() + datetime.timedelta(days=1),
     )
     return mockAnnotationJobParams2
+
+
+
+@pytest.fixture
+def mock_Extraction_AnnotationJobParams():
```
cakeinsauce, line 66: 1. make lowercase name
2. use typings to show what is return, no need for type in the name.

jobs/tests/test_db.py
```diff
@@ -151,6 +152,25 @@ def create_mock_annotation_job_in_db(
     return result
 
 
+def create_mock_annotation_extraction_job_in_db(testing_session, mock_Extraction_AnnotationJobParams):
```
cakeinsauce, line 155: we mock database and focus on unit tests. you should not save anything in there

cakeinsauce (changes requested):
PTAL

jobs/jobs/airflow_utils.py
```diff
@@ -33,9 +41,11 @@ def get_configuration():
 
 def get_api_instance():
     configuration = get_configuration()
-
-    with client.ApiClient(configuration) as api_client:
+    api_client = client.ApiClient(configuration)
+    try:
```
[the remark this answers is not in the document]
AliBugday: I converted it to a generator using yield; the previous version used a return statement. I also wanted to make sure the resources were closed.

jobs/jobs/utils.py
```diff
@@ -555,7 +558,11 @@ async def update_job_in_annotation(
 
 
 async def get_job_progress(
-    job_id: int, session: Session, current_tenant: Optional[str], jw_token: str
+    job_id: int,
+    session: Session,
+    current_tenant: Optional[str],
+    jw_token: str,
+    api_client: client.ApiClient,
 ) -> Optional[Dict[str, int]]:
```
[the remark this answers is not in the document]
AliBugday: The frontend expects the response in this format. The returned data should include the total and finished task values so it can display the job’s completion percentage. Since the response also includes the job mode, returning a dictionary with these values seems appropriate than other data structures.

jobs/jobs/utils.py
```diff
@@ -581,9 +584,137 @@ async def get_job_progress(
...
+    pipeline_completion = 0
+    if pipeline_status == AirflowPipelineStatus.success.value:
+        new_state = Status.finished
+        pipeline_completion = 1
+    elif pipeline_status == AirflowPipelineStatus.failed.value:
+        new_state = Status.failed
+        pipeline_completion = 0
+    elif pipeline_status == AirflowPipelineStatus.running.value:
+        new_state = Status.in_progress
+        pipeline_completion = 0
```
cakeinsauce, line 680: you already set pipeline_completion = 0 outside of if statements, no need to assign it here again

jobs/tests/test_utils.py
```diff
@@ -556,6 +561,126 @@ async def test_execute_pipeline_positive(jw_token):
...
+    [
+        (
+            AirflowPipelineStatus.success.value,
+            {"total": 1, "finished": 1, "mode": str(JobMode.Automatic)},
+        ),
+        (
+            AirflowPipelineStatus.failed.value,
+            {"total": 1, "finished": 0, "mode": str(JobMode.Automatic)},
+        ),
+    ],
```
cakeinsauce, line 578:
```suggestion
    ("dag_status", "expected_progress"),
    (
        (
            AirflowPipelineStatus.success.value,
            {"total": 1, "finished": 1, "mode": str(JobMode.Automatic)},
        ),
        (
            AirflowPipelineStatus.failed.value,
            {"total": 1, "finished": 0, "mode": str(JobMode.Automatic)},
        ),
    ),
```

jobs/tests/test_utils.py
```diff
@@ -556,6 +561,126 @@ async def test_execute_pipeline_positive(jw_token):
...
+            {"total": 2, "finished": 2, "mode": str(JobMode.Automatic)},
+        ),
+        (
+            JobMode.Automatic,
+            JobType.ExtractionWithAnnotationJob,
+            {"total": 3, "finished": 2},
+            {"total": 3, "finished": 2, "mode": str(JobMode.Automatic)},
+        ),
+    ],
+)
```
cakeinsauce, line 646:
```suggestion
@pytest.mark.parametrize(
    ("job_mode", "job_type", "fetch_result", "expected_progress"),
    (
        (
            JobMode.Manual,
            JobType.AnnotationJob,
            {"total": 2, "finished": 2},
            {"total": 2, "finished": 2, "mode": str(JobMode.Manual)},
        ),
        (
            JobMode.Manual,
            JobType.AnnotationJob,
            {"total": 3, "finished": 2},
            {"total": 3, "finished": 2, "mode": str(JobMode.Manual)},
        ),
        (
            JobMode.Automatic,
            JobType.ExtractionWithAnnotationJob,
            {"total": 2, "finished": 2},
            {"total": 2, "finished": 2, "mode": str(JobMode.Automatic)},
        ),
        (
            JobMode.Automatic,
            JobType.ExtractionWithAnnotationJob,
            {"total": 3, "finished": 2},
            {"total": 3, "finished": 2, "mode": str(JobMode.Automatic)},
        ),
    ),
)
```

jobs/tests/test_utils.py
```diff
@@ -556,6 +561,126 @@ async def test_execute_pipeline_positive(jw_token):
...
+async def test_get_job_progress_states(
+    jw_token: str,
+    monkeypatch: MonkeyPatch,
+    job_mode: JobMode,
+    job_type: JobType,
+    fetch_result: Dict[str, int],
+    expected_progress: Dict[str, any],
+):
```
cakeinsauce, line 654:
```suggestion
) -> None:
```

jobs/tests/test_utils.py
```diff
@@ -556,6 +561,126 @@ async def test_execute_pipeline_positive(jw_token):
...
+            {"total": 1, "finished": 0, "mode": str(JobMode.Automatic)},
+        ),
+    ],
+)
+async def test_get_extraction_job_progress_states(
+    jw_token: str,
+    monkeypatch: MonkeyPatch,
+    dag_status: AirflowPipelineStatus,
+    expected_progress: Dict[str, any],
+):
```
cakeinsauce, line 585:
```suggestion
) -> None:
```

cakeinsauce (changes requested):

cakeinsauce (approved):
Great job!

Issue #1005, opened 2025-03-26
Files are not shown when a job is created using revisions


halilaydin242188:
Recreation steps:
- start creating a job
- select revisions and go next
- select Human in the loop (manual job), fill the forms and create the job

Expected result:
- in the job review page, the 'Files' section list the files that are related to the revisions we selected

Actual result:
- No files are shown:

![Image](https://github.com/user-attachments/assets/fa616c88-3e39-4b70-9baf-ea50097a6f3f)

PR #1011, opened 2025-04-07
fix: files are not shown when a job is created using revisions

halilaydin242188:
closes: #1005

### DIFF
--- a/jobs/jobs/utils.py
+++ b/jobs/jobs/utils.py
@@ -1,13 +1,4 @@
-from typing import (
-    Any,
-    Dict,
-    Generator,
-    Iterator,
-    List,
-    Optional,
-    Tuple,
-    Union,
-)
+from typing import Any, Dict, Generator, Iterator, List, Optional, Tuple, Union
 
 import aiohttp.client_exceptions
 import fastapi.encoders
--- a/jobs/jobs/utils.py
+++ b/jobs/jobs/utils.py
@@ -898,12 +898,18 @@ async def validate_create_job_previous_jobs(
     return [j.id for j in previous_jobs]
 
 
-async def get_file_ids_of_revisions(
-    revisions: List[str], current_tenant: str, jwt_token: str
-) -> Optional[List[int]]:
-
+async def update_create_job_params_using_revisions(
+    job_params: schemas.JobParams, current_tenant: str, jwt_token: str
+) -> None:
     response = await get_annotations_by_revisions(
-        current_tenant=current_tenant, jw_token=jwt_token, revisions=revisions
+        current_tenant=current_tenant,
+        jw_token=jwt_token,
+        revisions=list(job_params.revisions),
     )
 
-    return list({data["file_id"] for data in response.get("data", [])})
+    unique_file_ids_of_revisions = set(
+        [data["file_id"] for data in response.get("data", [])]
+    )
+    job_params.files = list(
+        unique_file_ids_of_revisions.union(job_params.files)
+    )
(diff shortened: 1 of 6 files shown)


jobs/jobs/utils.py
```diff
@@ -788,6 +788,42 @@ async def get_annotation_revisions(
...
+    current_tenant: Optional[str], jw_token: str, revisions: List[str]
+) -> Optional[Dict[str, Any]]:
+    """Get annotations by filtering"""
+
+    headers = {
+        "X-Current-Tenant": current_tenant,
+        "Authorization": f"Bearer: {jw_token}",
+    }
+
+    timeout = aiohttp.ClientTimeout(total=5)
```
cakeinsauce, line 801: don't we have timeout config somewhere? let's use that.

jobs/jobs/utils.py
```diff
@@ -860,3 +896,20 @@ async def validate_create_job_previous_jobs(
...
+    job_params: schemas.JobParams, current_tenant: str, jwt_token: str
+) -> None:
+    response = await get_annotations_by_revisions(
+        current_tenant=current_tenant,
+        jw_token=jwt_token,
+        revisions=list(job_params.revisions),
+    )
+
+    unique_file_ids_of_revisions = set(
+        [data["file_id"] for data in response.get("data", [])]
```
cakeinsauce, line 911: what if payload changes? it will fail, let's add some processing here. at least check if file_id key is present, if not, skip

cakeinsauce (changes requested):
ptal

processing/processing/config.py
```diff
@@ -40,7 +40,7 @@ class Settings(BaseSettings):
     retry_attempts: int = 3
     retry_statuses: Set[int] = {501, 502, 503} | {i for i in range(505, 600)}
     delay_between_retry_attempts: int = 1  # in seconds
-    request_timeout: int = 3 * 60 * 60
+    request_timeout: int = 60  # in seconds
```
cakeinsauce, line 43: let's not touch it here, confusing to see in that pr

jobs/tests/test_utils.py
```diff
@@ -1193,3 +1194,71 @@ async def test_execute_external_pipeline(sign_s3_links: bool):
             )
         else:
             assert FakePipeline.calls[-1]["files"][0].get("signed_url") is None
+
+
+@pytest.mark.asyncio
```
cakeinsauce, line 1199: Remove this decorator and change this pytest config to `auto`
https://github.com/epam/badgerdoc/blob/2177c64d6ab8b1c8b34228ce207f1f64b9bde702/jobs/pytest.ini#L2
IDK why it was set to `strict` there (default behaviour)

jobs/jobs/utils.py
```diff
@@ -788,6 +786,39 @@ async def get_annotation_revisions(
...
+            method="POST",
+            url=f"{ANNOTATION_SERVICE_HOST}/annotation",
+            headers=headers,
+            body=post_data,
+        )
+    except aiohttp.client_exceptions.ClientError as err:
+        logger.exception("Failed request to get annotations by revisions")
+        raise fastapi.HTTPException(
+            status_code=fastapi.status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=f"Failed request to the Annotation Manager: {err}",
```
cakeinsauce, line 816: I think 422 is not a correct error code for that. I'd set it to 400. And also do not expose internal exception to outside. Consider http exception as an end user message. User has no idea what is an annotation manager.

cakeinsauce (changes requested):

cakeinsauce (approved):
great!

Issue #802, opened 2023-10-13
Add "Sort ascending/descending" functionality for files sorting 


KaterynaKovrizhnykh:
**Problem:**

Add "Sort ascending/descending" functionality so that user will be able to sort files by name or last modified date in ascending or descending order when he is on the "Documents" tab

**Solution:**

Add "Sort ascending/descending" functionality to "Sort by" drop-down for files on the "Documents" tab

PR #1014, opened 2025-04-11
feat: Add "Sort ascending/descending" functionality for files sorting  #802

SashaGraves:
Add "Sort ascending/descending" functionality to "Sort by" drop-down for files on the "Documents" tab

image


SashaGraves:
Probably need to change arrow color to gray. Black looks out-of-row

seliankinfedor (approved):
todo: we need to enhance this icon with hover/focus styles, because currently after "sort by" dropdown I do not know where am i

PR #1017, opened 2025-04-17
feat: enable enter key navigation in page picker

chgayane:
- Added controlled input state management for PickerInput component
- Implemented useEffect to sync input value with current page changes
- Added direct keyboard event handling via rawProps.input
- Included input validation for page number boundaries
- Maintained existing dropdown selection functionality

Fixes issue where enter key press didn't navigate to selected page

https://github.com/user-attachments/assets/da37bbc8-4385-499e-9a84-038788efea22

### DIFF
--- a/web/src/shared/components/document-toolbar/index.tsx
+++ b/web/src/shared/components/document-toolbar/index.tsx
@@ -34,10 +34,11 @@ export const DocumentToolbar: FC<TDocumentToolbar> = ({
     const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
         if (e.key === 'Enter') {
             const pageNumber = parseInt(pageInputValue, 10);
-            if (!isNaN(pageNumber)) {
-                const validPage = Math.min(Math.max(1, pageNumber), countOfPages);
-                onPickerValueChange(validPage);
-                setPageInputValue(String(validPage));
+            if (!isNaN(pageNumber) && pageNumber >= 1 && pageNumber <= countOfPages) {
+                onPickerValueChange(pageNumber);
+                setPageInputValue(String(pageNumber));
+            } else {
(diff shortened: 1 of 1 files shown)


web/src/shared/components/document-toolbar/index.tsx
```diff
@@ -70,6 +90,15 @@ export const DocumentToolbar: FC<TDocumentToolbar> = ({
                     getName={(item) => String(item)}
                     selectionMode="single"
                     disableClear={true}
+                    searchPosition="input"
+                    valueType="id"
+                    rawProps={{
+                        input: {
+                            defaultValue: pageInputValue,
```
seliankinfedor, line 97: what for we need default value here?

web/src/shared/components/document-toolbar/index.tsx
```diff
@@ -70,6 +90,15 @@ export const DocumentToolbar: FC<TDocumentToolbar> = ({
                     getName={(item) => String(item)}
                     selectionMode="single"
                     disableClear={true}
+                    searchPosition="input"
+                    valueType="id"
```
seliankinfedor, line 94: looks like "id" is already defined as default value here
image

web/src/shared/components/document-toolbar/index.tsx
chgayane: I agree, the defaultValue in rawProps.input is redundant and potentially conflicting with state management. I think it is better to remove it and keep id.

web/src/shared/components/document-toolbar/index.tsx
chgayane: The valueType="id" is still needed to ensure the PickerInput handles primitive number values correctly.

seliankinfedor (approved):

Issue #777, opened 2023-10-13
Non-admin user can access files and tabs via direct link


KaterynaKovrizhnykh:
**Steps to reproduce:**

Login as a non-admin user
Try to open http://badgerdoc-dev.westeurope.cloudapp.azure.com:8083/categories (f.e.) or any other page/tab that is shown only for admin users

**Expected result:**
Non-admin user should not have access to these pages 

**Actual result:**
Non-admin user can open and navigate through these pages can perform annotation/validation task that weren't assigned to him:

PR #1025, opened 2025-05-07
fix: Non-admin user can access files and tabs via direct link #777

SashaGraves:
Steps to reproduce:
Login as a non-admin user
Try to open http://badgerdoc-dev.westeurope.cloudapp.azure.com:8083/categories (f.e.) or any other page/tab that is shown only for admin users

Expected result:
Non-admin user should not have access to these pages

Expected result is met.

A non-admin user no longer has access to modify or annotate items they shouldn’t have access to. The only change they can make is creating a new Category. If it is not a requirement, it’s up to the backend to decide whether this endpoint should be restricted for non-admins. 

This pr blocks the Category route on the frontend side — all other routes are working fine.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -28,7 +28,8 @@ import {
     JOBS_PAGE,
     MODELS_PAGE,
     PIPELINES_PAGE,
-    REPORTS_PAGE
+    REPORTS_PAGE,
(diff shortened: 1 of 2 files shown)


chgayane (approved):

rNiall (approved):
LGTM

PR #1026, opened 2025-05-07
feat: Ensure selected files are saved when switching the view from Grid to List #1030

chgayane:
Task: Persist Selected Files Across Grid and List View Switches in DocumentsTableConnector
closes: [https://github.com/epam/badgerdoc/issues/1030](https://github.com/epam/badgerdoc/issues/1030)

image
image


web/src/connectors/documents-table-connector/documents-table-connector.tsx
```diff
@@ -94,17 +95,17 @@ export const DocumentsTableConnector: React.FC<DocumentsTableConnectorProps> = (
...
     useEffect(() => {
         if (fileIds) {
+            setSelectedFiles(fileIds);
             onTableValueChange({
                 ...tableValue,
                 checked: fileIds
             });
         }
-    }, []);
+    }, [fileIds]);
```
rNiall, line 104: Did you check any side effects of doing refresh on fileIds change? 
Just curious why we change from one time refresh -> to any fileIds change.

web/src/shared/helpers/utils.ts
```diff
@@ -148,3 +148,8 @@ export const useOutsideClick = (ref: any, callback: () => void) => {
         };
     });
 };
+
+export const arraysEqual = (a: number[], b: number[]): boolean => {
+  if (a.length !== b.length) return false;
+  return a.every((value, index) => value === b[index]);
```
rNiall, line 154: Are we sure about the order of "b" array.
We also have lodash library installed. `isEqual(a, b)`

web/src/connectors/documents-table-connector/documents-table-connector.tsx
chgayane: The fileIds is essential for initializing and updating the selection state (`selectedFiles` and `tableValue.checked`) when the parent component passes new file IDs. Without reacting to fileIds changes, selections would not work or update correctly when switching between Grid  and List views.

web/src/shared/helpers/utils.ts
chgayane: I’ll switch `arraysEqual` to use Lodash’s `isEqual` for future-proofing, as it can handle cases where order might matter.

rNiall: Unfortunately it's not
Screenshot 2025-05-14 at 13 41 37

chgayane: In these cases, `checked` and `selectedFiles` have the same order, so `isEqual` works correctly without triggering unnecessary updates. I haven’t yet seen cases where they differ in order (e.g., `[3, 4]` vs. `[4, 3]`), which could cause updates due to `isEqual`’s order sensitivity. Based on the code, if they do differ, `selectedFiles` is updated to match `checked`, ensuring the order remains consistent and doesn’t break functionality.

rNiall (approved):

PR #1027, opened 2025-05-12
fix: pydantic v2 related issues

halilaydin242188:
This PR includes a workaround in addition to the standart version upgrade.

Context:
Starting from FastAPI v0.109.0 , the applications that uses "root_path" param in FastAPI instantiation is affected.
The routes that start with the same text as the 'root_path' value return "Not Found" response.
Example: root_path= "/users" , route= "/users/{user_id}"

The related issue links:

Workaround solution(using uvicorn 'root_path' param) is found in here:
https://github.com/fastapi/fastapi/discussions/11015
Similar issue:
https://github.com/fastapi/fastapi/discussions/9018
As recommended in the first link, uvicorn '--root-path' param in dockerfile is used to resolve this issue.
Also for api tests, the 'root_path' of app is removed before TestClient instantiation.

These workarounds can be removed once the issue is officially fixed in FastAPI.


jobs/jobs/utils.py
```diff
@@ -578,14 +578,19 @@ async def get_job_progress(
...
+            method="GET", url=url, headers=headers
+        )
     except aiohttp.client_exceptions.ClientError as err:
         logger.exception(f"Failed request url = {url}, error = {err}")
         raise fastapi.HTTPException(
             status_code=fastapi.status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail=f"Failed request to the Annotation Manager: {err}",
         )
 
+    if req_status_code == 404:
```
halilaydin242188, line 591: annotation service put job api returning 404 case is missed so that was causing error

cakeinsauce: Do we have a processing of this none anywhere? additionally, the return typing is incorrect for the function right now

halilaydin242188: yes, the only caller of this function which is get_job_progresss router was already processing none values. this function was checking if job is not found in the db but it is not checking if annotation can find the job in its db.  Also the return type is Optional and this func was returning none already, do i need to change it ?

cakeinsauce (approved):

cakeinsauce (approved):

PR #1049, opened 2025-07-31
fix: MainMenu item collapsing and overflow behavior

Svetlana-Sat:
Refactored MainMenu rendering logic to use AdaptiveItemProps with explicit priority and collapsedContainer settings. 
This enables proper collapsing of items into the "More" and burger menus.
Previously, some items were missing from the overflow menu, and the burger menu appeared empty due to lack of hiddenItems handling.
Now all menu items are passed via items prop, allowing MainMenu to correctly manage visibility and layout.


seliankinfedor:
for big zoom values when we hide settings inside menu and trying to expand them, they overlay existing menu items
image
image
so categories and reports are hidden
image

seliankinfedor (approved):

PR #1084, opened 2026-04-28
[Chat] Fix: Context reference is not converted to chip without trailing slash

### LINKED ISSUE #1082: [Bug][Chat] Context reference is not converted to chip without trailing slash
## Description
Currently, context reference conversion works inconsistently depending on whether the reference has a trailing slash.

For example, document reference is converted to a chip only when the path ends with `/`:

- Works: `{{/badgerdoc/document/7/}}`
- Does not work: `{{/badgerdoc/document/7}}`

This behavior should be consistent for all supported context entities.

## Expected Result
Context references should be recognized and converted to chips both with and without a trailing slash.

Examples:

- `{{/badgerdoc/document/7}}`
- `{{/badgerdoc/document/7/}}`
- `{{/badgerdoc/document/7/page/1}}`
- `{{/badgerdoc/document/7/page/1/}}`
- block/extraction references with and without trailing `/`

## Actual Result
Some references are converted only when they contain a trailing `/`.

## Requirements
- [ ] Update reference parsing to support optional trailing slash.
- [ ] Apply this behavior consistently for all supported entities:
  - document
  - page
  - block / extraction reference
- [ ] Ensure chip conversion works for manually typed references.
- [ ] Ensure chip conversion works for references inserted via UI actions.
- [ ] Add or update tests for both variants:
  - with trailing `/`
  - without trailing `/`

## Acceptance Criteria
- [ ] `{{/badgerdoc/document/7}}` is converted to a chip.
- [ ] `{{/badgerdoc/document/7/}}` is converted to a chip.
- [ ] All other supported context references work with and without trailing `/`.
- [ ] Existing behavior for escaped references remains unchanged.

grustnenkiy:
https://github.com/epam/badgerdoc/issues/1082

### DIFF
--- a/web/frontend/src/features/workspace/components/prompt-context-editor.test.tsx
+++ b/web/frontend/src/features/workspace/components/prompt-context-editor.test.tsx
@@ -13,10 +13,43 @@ import {
 import type { PromptContextPathInserter } from '@/features/workspace/helpers/extraction-chat-context'
 
 describe('prompt context editor serialization', () => {
-  it('turns valid context links into token nodes', () => {
-    const content = createPromptEditorContent(
-      "Summarize {{/badgerdoc/document/123/extraction/456/page/1/(//div[@id='block_1_1'])}}"
-    )
+  it.each([
+    {
+      prompt: 'Summarize {{/badgerdoc/document/123}}',
+      path: '/badgerdoc/document/123',
+    },
+    {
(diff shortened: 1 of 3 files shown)


sem0ark (approved):
Checked all regexes with regex101, everything works, tested locally as well.

Screenshot 2026-04-30 at 11 48 50
Screenshot 2026-04-30 at 11 47 05

PR #1092, opened 2026-04-30
fix(osd): disable default toolbar to stop 404 image requests

### LINKED ISSUE #1079: [Bug][OSD] Check and fix missing OpenSeadragon toolbar asset requests
### Summary
OpenSeadragon image button assets are requested by the application but return `404 Not Found`.

### Description
Several default OpenSeadragon toolbar assets are being requested on page load:

- `home_rest.png`
- `home_grouphover.png`
- `home_hover.png`
- `home_pressed.png`
- `fullpage_rest.png`
- `fullpage_grouphover.png`

It is not clear whether these assets are actually required by the current UI. This should be checked before adding them back to the build/static assets.

### Expected Result
Either:
- the assets are required and should load successfully, or
- the assets are not used and should not be requested at all.

### Actual Result
The assets are requested and return `404 Not Found`.

### Notes
Please verify whether default OpenSeadragon toolbar buttons/images are still needed in the current viewer implementation. If they are not needed, the configuration should be adjusted to prevent these requests.

katsiarynaby:
OpenSeadragon was requesting default toolbar image assets `(home_*.png, fullpage_*.png)`, which returned 404.

The application uses custom toolbar controls, so default navigation controls are not needed.

Disabled default navigation UI via `showNavigationControl: false` to prevent unnecessary asset requests.

### DIFF
--- a/web/frontend/src/components/collection-viewer/config.ts
+++ b/web/frontend/src/components/collection-viewer/config.ts
@@ -6,6 +6,7 @@ export const NEXT_BTN_ID = 'viewer-next'
 export const PREV_BTN_ID = 'viewer-prev'
 
 export const DEFAULT_OSD_CONFIG: Options = {
+  showNavigationControl: false,
   // toolbar settings
   zoomInButton: ZOOM_IN_BTN_ID,
   zoomOutButton: ZOOM_OUT_BTN_ID,


sem0ark (approved):
Tested on local deployment, no 404 found approved.

PR #1094, opened 2026-05-01
Fix: use current extraction id for block context paths

### LINKED ISSUE #1096: [Chat] Newly created extraction block uses stale extraction ID in context path
## Summary

When a new extraction block is created and immediately added to the context, the generated path contains the ID of the previously created extraction block instead of the newly created one.

The issue disappears after a page refresh — after reloading, the same block produces the correct path with the expected extraction ID.

This suggests that stale extraction state, cached references, or delayed state synchronization may be used during context generation.

## Steps to Reproduce

1. Open a document.
2. Create a new extraction block.
3. Without refreshing the page, add the newly created block to the prompt context.
4. Inspect the generated block path.

## Actual Result

The generated path contains the ID of the **previously created extraction block**, not the newly created one.

Example:

```text
Expected: {{/badgerdoc/document/{document_id}/extraction/{new_extraction_id}/...}}
Actual:   {{/badgerdoc/document/{document_id}/extraction/{previous_extraction_id}/...}}

grustnenkiy:


khyurri:
This PR closes: https://github.com/epam/badgerdoc/issues/1096
Tested, working good!

PR #1109, opened 2026-05-07
[TipTap] Fix: Accept action does not save the first extraction for documents without existing extractions

### LINKED ISSUE #1098: [TipTap] Accept action does not save the first extraction for documents without existing extractions
## Summary

When working with a document that does not have any existing extraction blocks, creating a new extraction block works in the UI, but clicking **Accept** does not trigger persistence.

As a result, the newly created extraction is not saved.

This issue appears only for documents with an initially empty extraction list.

## Steps to Reproduce

1. Open a document that has **no existing extractions**.
2. Create a new extraction block.
3. Click **Accept**.

## Actual Result

- Clicking **Accept** does not produce any visible effect.
- The extraction is not saved.
- No state update or persistence happens.
- After page reload, the newly created extraction is missing.

## Expected Result

Clicking **Accept** should persist the newly created extraction, even when it is the first extraction in the document.

grustnenkiy:


khyurri:
Tested, works perfect!

Issue #1095, opened 2026-05-04
Create build script to run version locally


AlexeyGirin:

khyurri:
@AlexeyGirin , can you please fill the description and requirements for this task?

sem0ark:
Draft of the description:

As a part of goals for improving the testing in scope of the project development, it was asked to develop scripts for automatic export or setup of data related to a given document, which should allow fast reproducibility of common scenarios.

AC:
- There is a fast way to set up multiple documents with a predefined configuration, such as uploaded document with DZI and a couple of extraction results.
- There is a fast way to manually set up a state for a given document and export it into a shareable format such as a ZIP archive to store/share separately.

sem0ark:
Specified as not needed by @khyurri 

Closing the issue.

PR #1110, opened 2026-05-07
Add export/import scripts

sem0ark:
#1095

In scope of the following PR two new scripts are added to quickly setting up and/or exporting specific configurations of extractions and documents to avoid every time manually set up the state. This should allow fast manual PR and/or regression testing.

Added two new Makefile commands to automatically handle upload/download from containers.


sem0ark:
Decided to not be beneficial, closing the PR.

PR #1127, opened 2026-05-20
Add PNG rendering fallback when DZI tile assets are unavailable

### LINKED ISSUE #1107: [OSD] Fallback to PNG rendering in OpenSeadragon when tiled assets are unavailable
## Description

Currently, the document viewer expects tiled image assets (DZI / tile source) to be available for page rendering in OpenSeadragon.

If tile metadata or tiled assets are missing, the page cannot be displayed.

We need to introduce a fallback rendering mode that allows the viewer to display a regular PNG image using OpenSeadragon's image source (`{ type: 'image', url }`) when tiled rendering is unavailable.

At this stage, the PNG fallback is intended for **view-only mode**.

Interactive features that depend on tiled rendering (such as selection, region highlights, annotation overlays, extraction block interaction, etc.) are **not required** for PNG fallback in this scope.

## Requirements

- [ ] Use tiled rendering as the primary rendering mode when tile source data is available.
- [ ] If tiled rendering cannot be initialized or fails because tile source data or tiled assets are missing, fall back to PNG rendering.
- [ ] The fallback can be triggered based on available metadata, API response, or OpenSeadragon load/init failure handling — the exact detection strategy can be defined during implementation.
- [ ] If fallback is triggered, open the page in OpenSeadragon using:

```ts
{
  type: "image",
  url: pagePngUrl,
}
```

- [ ] PNG fallback should render the page without errors or viewer crashes.
- [ ] Basic viewer interactions should remain available:
  - zoom
  - pan
  - fit-to-screen

- [ ] Selection and interactive overlays are **not required** in PNG fallback mode:
  - area selection
  - OCR block selection
  - extraction block overlays
  - highlight editing

- [ ] The viewer should gracefully switch to read-only display mode when PNG fallback is used.

## Acceptance Criteria

- [ ] A page with available tiles behaves exactly as before.
- [ ] If tiled rendering cannot be initialized, the viewer attempts PNG fallback automatically.
- [ ] A page without working tiles is still visible in the viewer via PNG fallback.
- [ ] No runtime errors occur when tile metadata or tiled assets are missing.
- [ ] The viewer remains zoomable and pannable in fallback mode.
- [ ] Interactive selection tools are disabled or hidden in fallback mode.

katsiarynaby:
## Summary

The document viewer relies on tiled DZI rendering through OpenSeadragon. When a document has no DZI assets (conversion didn't run, partially failed, or files in storage are missing/broken) the viewer cannot display any pages.

This PR introduces a PNG fallback that pulls the per-page PNG renditions from the existing `/document/{id}/renditions/` endpoint and feeds them into OpenSeadragon's simple-image source. When fallback is active the viewer operates in a read-only mode — overlays, area selection, and edit mode are disabled, but zoom / pan / fit-to-screen work as before.

## How it works

Two-tier detection so both data-level and runtime-level failures are covered:

1. **API-response check (primary).** `getDocumentPages` calls `/dzi/` first.
   If the response is empty or the call throws, it falls back to fetching  `/renditions/`, sorts the result by `metadata.page` ascending, and returns each PNG as `{ type: 'image', url }`.
2. **OSD runtime check (safety net).** `useOsdViewer` listens for OpenSeadragon's `open-failed` event. When DZI URLs are returned by the API but the underlying assets fail to load (e.g. files missing from MinIO), `CollectionViewer` fetches `/renditions/` and rebuilds the viewer with PNG sources. A ref guards against re-triggering for the same document.

A `PageSource` discriminated union (`{ type: 'dzi' | 'image', url }`) is threaded through the service layer, adapter contracts, and the React Query hook. `useOsdViewer` unwraps each entry to the appropriate OpenSeadragon tileSource shape (`string` for DZI, `{ type: 'image', url }` for PNG).

When `effectivePages` contains only image entries, the viewer enters read-only mode:
- The edit-mode pencil button is hidden via `canUseEditMode={false}`.
- `useExtractionHighlights` receives empty overlay items, null selection, and an empty created-id set, so no overlays are drawn and no MouseTrackers are attached.
- The rubberband-draw `useEffect` returns early because `isEditMode` is forced false.
- Pan stays enabled; the crosshair cursor never appears.

### DIFF
--- a/web/frontend/src/components/collection-viewer/collection-viewer.tsx
+++ b/web/frontend/src/components/collection-viewer/collection-viewer.tsx
@@ -96,12 +96,16 @@ export function CollectionViewer({
     (effectivePages?.length ?? 0) > 0 && effectivePages!.every((p) => p.type === 'image')
   const effectiveCanUseEditMode = canUseEditMode && !isPngFallback
   const effectiveIsEditMode = isEditMode && !isPngFallback
+
+  const emptyOverlays = useMemo<Record<number, OverlayBox[]>>(() => ({}), [])
+  const emptyCreatedIds = useMemo<Set<string>>(() => new Set(), [])
+
   useExtractionHighlights({
     viewer,
-    overlayItems: isPngFallback ? {} : highlights,
+    overlayItems: isPngFallback ? emptyOverlays : highlights,
     selectedHighlightId: isPngFallback ? null : activeHighlightId,
     isEditMode: effectiveIsEditMode,
-    createdHighlightIds: isPngFallback ? new Set<string>() : createdHighlightIds,
+    createdHighlightIds: isPngFallback ? emptyCreatedIds : createdHighlightIds,
     onHighlightClick,
     onHighlightUpdate,
     onHighlightCreate,
--- a/web/frontend/src/components/collection-viewer/collection-viewer.tsx
+++ b/web/frontend/src/components/collection-viewer/collection-viewer.tsx
@@ -1,15 +1,16 @@
 import { ViewerToolbar, type PageChatContext } from '@/components/collection-viewer/viewer-toolbar'
 import { useViewerNavigation } from '@/components/collection-viewer/use-viewer-navigation'
 import { useOsdViewer } from '@/components/collection-viewer/use-osd-viewer'
-import { OverlayBox } from '@/shared/api/badgerdoc/types'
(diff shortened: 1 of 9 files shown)


seliankinfedor (approved):

katsiarynaby:
@khyurri please check and confirm that all is working as you expected

